PR #1, opened 2022-01-27
fill in ACE1pack

gelzinyte:
A draft of a complete package. @casv2 (and anyone else if they wish!) could you please leave some early feedback on the work in progress? Anything missing, any suggestions? Some things need better names (e.g. `everything.jl` & `fit_ace`). 

Meanwhile, I'll work on debugging, tests & CI, TODO's and docstrings in the code. These should be done before merging, but I have left some "enhancement" comments in the code that might be useful in slightly longer term.


cortner:
Thank you, Elena! We need a brief discussion how to go about merging yours vs adapt to latest IPFitting first; cc @casv2 ; cf #2 ; I see you've already introduced that version bound though so maybe you are further with this than I thought.

casv2:
Yes, Elena and I have been coordinating quite closely in designing ACE1pack and IPFitting, and will continue to do so. Let's discuss tomorrow, I think we're quite close. Regarding the regularisation interface the idea is to put the ACE1.scaling oneliner into ACE1pack. This way we set up the preconditioning matrix in ACE1pack (which has the ACE dependency) and send it to IPFitting for fitting.

cortner:
I'm ok with that.

gelzinyte:
The TiAl tutorial now works - see `test_fit.jl` (which is more of a example/draft than an actual unit test, for now). 

I'm working on the unit tests, but that can also done as another pull request if you think that it's better to merge and open this up for everyone as soon as possible. 

Looking forward to any comments on the code!

cortner:
so in principle you think it is ready to tag a first 0.0.1 version? and let people try it out?

gelzinyte:
yes, I think so!

gelzinyte:
The current tests mostly just call functions to make sure they execute without an error. I thought anything less trivial (mainly - that the objects constructed from given parameters are correct) would be tested within the underlying modules and that testing for specific outputs (e.g. that default values are as expected) seemed maybe excessive. Is there anything else that would be good to test for? 

The tests on GitHub currently fail, has anyone stumbled upon anything similar? 
- On the x86 arch only: HDF5 isn't precompiled correctly
- Tests on x64 arch do finish, but the errors of a TiAl fit are different from those I get when running the tests locally. 

I think this should have everything that is needed at the first instance. I would eventually like to use multitransform and be able to specify polynomial degree for each correlation order, but I would first like to integrate ACE1pack with Workflow.

.github/workflows/CI.yml
```diff
@@ -29,8 +31,14 @@ jobs:
...
         with:
-          version: ${{ matrix.version }}
+          version: ${{ matrix.julia-version }}
           arch: ${{ matrix.arch }}
+      - uses: actions/setup-python@v2
+        with: 
+          python-version: ${{ matrix.python-version }}
+      # - name: install Python packages
+      #   run: |
+      #     pip install ase scikit-learn
```
gelzinyte: ASE wasn't being installed automatically, but on a second thought maybe because Python wasn't present? Should the user install ASE in their Python environment before ACE1pack, or should it be installed with IPFitting, etc?

gelzinyte: It looks like ASE has to be installed after all

src/basis.jl
```diff
@@ -124,12 +177,14 @@ end
 TODO: needs docs 
 """ 
 function radbasis_params(; 
-         r0 = 2.5, 
-         rcut = 5.0,
-         rin = 0.5 * r0,
-         pcut = 2,
-         pin = 2
-         )
+      rin = nothing,
```
gelzinyte: I've changed the default, because I think it should be different for pair and rpi basis. Smooth inner cutoff at `0.5 * r0` for rpi and no inner cutoff for pair. Please correct me if I'm wrong! The defaults are now set at `rpi_basis_params` and `pair_basis_params`.

src/basis.jl
```diff
@@ -2,11 +2,14 @@
 # ------------------------------------------
 #   ACE Basis  
 
-export basis_params, degree_params, radbasis_params, transform_params
+export rpi_basis_params, pair_basis_params, degree_params, radbasis_params, transform_params
```
gelzinyte: Is there a Julia's rule of thumb on what should get exported and what should be called as `Module.function()`?

cortner: we can choose our own rules, generally speaking anything user oriented and which doesn't clash with other packages should be exported I think.

cortner: all the above look ok to me.

cortner:
sorry for being slow to wrap this up, but I see you are still pushing commits. I'll start reviewing the code a bit.

src/basis.jl
```diff
@@ -18,14 +21,14 @@ the argument is required.
 * `r0 = 2.5` : rough estimate for nearest neighbour distance
 * `basis1p = basis1p_params(; r0 = r0)` : one-particle basis parameters; cf `?basis1p_params` for details 
 * `transform = transform_params(; r0 = r0)` : distance transform parameters; cf `?transform_params()` for details
-* `degreetype = degree_params()` : class of sparse polynomial degree to select the basis; see `?degree_params` for details 
+* `degree = degree_params()` : class of sparse polynomial degree to select the basis; see `?degree_params` for details 
 """
-function basis_params(; 
+function rpi_basis_params(;
```
cortner, line 44: I'm not too sure about this renaming. My thought was to not be tied down by potentially. bad choices I made a long time ago. E.g. if `basis` is too generic, then maybe `ace_basis` would be a better choice here?

src/basis.jl
cortner: I see the point, but I would probably organize this differently. Arguably the pair basis IS the rad_basis, so you shouldn't need pair_basis  AND radbasis. I would go back to `rad_rpi_basis -> radbasis` and incorporate the current `rad_basis` into `pair_basis`.

src/read_params.jl
```diff
@@ -0,0 +1,54 @@
+
+# ------------------------------------------
+#   Read ACE fit parameters from file 
+
+using JSON, YAML, ACE1pack
```
cortner: I generally prefer to use the interface for reading, writing JSON that is implemented in JuLIP so that we can swap around the backend as needed and to not duplicate any work especially if we start making it more user-friendly then it should be done in only one place. Leave for now, but file an issue to revisit this please?

src/solver.jl
```diff
@@ -0,0 +1,72 @@
+
+# ------------------------------------------
+#   solver
+
+using IPFitting, ACE1, LinearAlgebra, JuLIP
+
+export solver_params, precondition_laplacian, apply_preconditioning!
+
+"""TODO documentation"""
+function solver_params(;
```
cortner: This is a construction that I don't like. I would write this as 
```Julia 
__solvers_params = Dict("QR" => _qr_params, "RRQR" => _rrqr_params)  # etc

function solver_params(; solver = nothing, kwargs...)
    @assert solver in keys(__solvers_params)
    return __solvers_params[solver](;kwargs...)
end

_qr_params() = Dict("solver" => "QR")

_rrqr_param(; tol = 1e-5) = Dict("solver" => "RRQR", "tol" => tol)

# etc 
```

cortner (changes requested):

cortner:
I've done a first pass - You've done an enormous amount of work, thank you!! 

One thing that isn't in my review but we really need to address this is that we should not store any data files in these repositories. They should be managed as artifacts.  [see the docs here](https://pkgdocs.julialang.org/v1/artifacts/)

gelzinyte:
###

src/basis.jl
gelzinyte: Following `ACE1.utils.jl`, we need to have `RPI_basis` (`rpi_basis`) and `PolyPairBasis` (`pair_basis`) and both need `TransformedPolys` (`rad_basis`) for the radial part. So even though pair basis is effectively a radial basis, we still need all three (that is, can't use `pair_basis` in place of `rad_basis`), because `pair_basis` turn `rad_basis`/`TransformedPolys` into `PolyPairBasis`. 

I was inconsistent with where from/how `rpi_basis` and `pair_basis` get the `rad_basis` - the former currently calls `generate_rpi_radbasis` from `generate_rpi_basis` and the later constructs `radbasis` directly in `generate_pair_basis`. To be consistent, I'll make both of them use `generate_radbasis`  (renamed from what is currently `generate_rpi_radbasis`). Maybe that's exactly what you meant? 

Somewhat more removed - I considered if it's possible or worth not to separate `basis` into `rpi_basis` and `pair_basis`. Then, to have the same behaviour as now, `ace_fit` parameters would have an entry: `fit_ace_params["basis"] = [rpi_basis_params, pair_basis_params]`. The parameters would get transformed into `RPI_basis` and `PolyPairBasis` by the same `generate_basis`, depending on the the parameters (e.g. a `type` field or parameters with `N=1` -> `PolyPairBasis` and `N>1` -> `RPI_basis`), and `IPSuperBasis` would be constructed from`[RPI_basis, PolyPairBasis]` like now. On the upside, it would be straightforward to mix and match what basis get combined into `IPSuperBasis` and add new ones without changing the user-facing behaviour. On the other hand, if `RPI_basis` + `PolyPairBasis` is all that is going to be used in ACE1, then I think the current slightly more explicit/verbose setup is more transparent and less error-prone.

src/read_params.jl
gelzinyte: Opened issue #3 .

.github/workflows/CI.yml
cortner: I don't remember whether I ever got this to work on GitHub CI, I would say this is not the most urgent, and we can focus on that later.

src/basis.jl
cortner: I like the idea of putting `pair_basis` into `ace_basis`, but I see no reason to have `pair_rad_params` and `rpi_rad_params`. There should simply be 
* `pair_params` : parameters for pair basis 
* `rad_params` : parameters for radial basis for RPI basis.

What case do you have in mind that this doesn't cover?

src/solver.jl
gelzinyte: That's much neater, updated in [37e3e22](https://github.com/ACEsuit/ACE1pack.jl/commit/37e3e2227a1600343ff239a0304c570c819b185f)

src/basis.jl
gelzinyte: I think you are suggesting this: 

```
function pair_params(species, maxdeg, r0, rin, pin, rcut, pcut, transform)
function rpi_params(species, N, maxdeg, r0, rad_basis, transform, degree)
function rad_params(rin, pin, rcut, pcut)
```

I am suggesting this: 
```
function pair_params(species, maxdeg, r0, rad_basis, transform)
function rpi_params(species, N, maxdeg, r0, rad_basis, transform, degree)
function rad_params(rin, pin, rcut, pcut)
```
both `rpi_params`/`generate_rpi_basis` and `pair_params`/`generate_pair_basis` would be using the same `rad_params`/`generate_rad_basis`, but would have different default values for `rin`& `pin` in `rad_params`.

src/basis.jl
gelzinyte: Okay! May I suggest exporting `*_params()` and `generate_*()`? I think both a convenient way of generating the parameters and making stuff from them are equally useable.

src/basis.jl
cortner: What you are suggesting makes sense in the context of the ACE1 code, but I thought it would be inconvenient for the user. But in the end you are the user and not I, so if you prefer this structure, then let's leave it like this. 

On the other point - I still like your suggestion to put pair_basis into ace_basis. I think this is worthwhile.

src/basis.jl
cortner: definitely export `*_params`. Why export `generate_*`? isn't that an internal thing? I would even go so far as to not document this since it gives us more freedom to break compatibility.

cortner: again as usual - if there is a good reason you as a user want this exported, I won't stand in the way. I'm just asking questions to make sure we are thinking this through.

gelzinyte: No, you're right, there shouldn't be a need to mess with any of the `generate_*`.

src/basis.jl
gelzinyte: Ah, I see what you meant in the original comment, I was thinking about it with ACE1 code in mind. 

I'll incorporate `pair_basis` into `ace_basis`. 

I didn't think `basis` was too generic and I could interpret `ace_basis` to mean both `rpi_basis` only and `rpi_basis` & `pair_basis` (which one is correct in general?). Also, currently there already are `fit_ace` and `ace_params`, so `basis` might be clear enough? What do you think?

src/read_params.jl
```diff
@@ -0,0 +1,45 @@
+
+# ------------------------------------------
+#   Read ACE fit parameters from file 
+
+using ACE1pack
+
```
gelzinyte, line 6: Following up on the discussion in #3, one actually doesn't need to use any of JuLIP dictionary reading here, that should be done by the user, e.g. 

```
using JuLIP
fit_params = load_dict(@__DIR__() * "/files/fit_params.json")
```
Since that's a dictionary already and dictionaries, not composite types are used throughout ACE1pack, there's no need to define a `read_dict` here. 

So what's left is a `fill_defaults!` utility function, which I think acts differently enough from `read_dict` to be named distinctly. 

I think the most common way of loading the parameters ahead of fitting will be
```
using ACE1pack, JuLIP
fit_params = fill_defaults!(load_dict(@__DIR__() * "/files/fit_params.json"))
fit_ace(fit_params)
```

cortner: Hm ... are you saying you want to read a dictionary with partial parameters and then fill in the defaults? Ok, I should have anticipated the need for this. Let me sit on this for a moment please.

src/basis.jl
cortner: I'm totally ok with just `basis` especially if it includes both the ACE and the pair basis.

.github/workflows/CI.yml
cortner: maybe turn off Github actions for now and file an issue?

src/basis.jl
cortner: I think we now agreed on just `basis`?

src/read_params.jl
cortner: Ok, I think I know how to do some of this: 
```Julia
# this function just shows the kwargs so you can see how the code works.
f(; kwargs...) = (@show kwargs)
```
e.g., 
```
julia> f(;a = 1, b = 2)
kwargs = Base.Pairs(:a => 1, :b => 2)
pairs(::NamedTuple) with 2 entries:
  :a => 1
  :b => 2
```

In principle you can splat a dictionary using `D...` to turn it into kwargs, but the problem is that the keys are strings and not symbols. E.g., 
```Julia 
julia> D = Dict("a" => 1, "b" => 2); f(; D...)
ERROR: TypeError: in typeassert, expected Symbol, got a value of type String
Stacktrace:
 [1] merge(a::NamedTuple{(), Tuple{}}, itr::Dict{String, Int64})
   @ Base ./namedtuple.jl:293
 [2] top-level scope
   @ REPL[49]:1
```

But we can just convert them (recursively): 
```Julia
_makesymbol(x) = x   # default 
_makesymbol(p::Pair) = (Symbol(p.first) => _makesymbol(p.second))  # recursion 
_makesymbol(D::Dict) = Dict(_makesymbol.([D...])...)    # special case dictionary 
```
e.g. 
```Julia
julia> _makesymbol(Dict("a" => 1, "b" => 2))
Dict{Symbol, Int64} with 2 entries:
  :a => 1
  :b => 2
```
and another example with nested dictionaries 
```Julia
julia> _makesymbol(Dict("a" => 1, "b" => 2, "c" => Dict("d" => 3)))
Dict{Symbol, Any} with 3 entries:
  :a => 1
  :b => 2
  :c => Dict(:d=>3)
```

And now we can splat the kwargs: 
```Julia
julia> D = Dict("a" => 1, "b" => 2); f(; _makesymbol(D)...)
kwargs = Base.Pairs(:a => 1, :b => 2)
pairs(::NamedTuple) with 2 entries:
  :a => 1
  :b => 2
```

cortner: As long as the Dict that the user passes can be passed into a single function call which then sorts out these parameters, this might "just work", but needs to be carefully tested of course.

cortner: I wonder though, should we move this to the next PR, so we can wrap this up and tag it?

scripts/generate_artifacts.jl
```diff
@@ -0,0 +1,59 @@
...
+        tarball_hash = archive_artifact(hash_, joinpath(tarfile))
+        bind_artifact!(artifacts_toml, label, hash_,
+                    download_info = [ (url, tarball_hash) ], lazy=true, force=true)
+    end
+end
+
+
+label = "TiAl_tiny_dataset"
+zipname = "TiAl_tiny.zip"
+url = "https://github.com/gelzinyte/test_julia_artifacts/blob/main/$(zipname)?raw=true"
```
gelzinyte, line 31: will update once/if https://github.com/ACEsuit/ACEData/pull/1 is merged

src/basis.jl
gelzinyte: Yes, the exported function is now `basis_params` only.

src/read_params.jl
gelzinyte: Thanks a lot for the suggestion!! Yes, I would rather not delay too much - have just opened issue #4 and moved your suggestion there.

src/read_params.jl
```diff
@@ -0,0 +1,46 @@
...
+export  fill_defaults!
+
+""" Recursively updates nested dictionaries with default parameters"""
+function fill_defaults!(params::Dict; param_key = "fit_params")
+    # Go through the nested dictionaries filling in the default values
+    params = _fill_default(params, param_key)
+    for (key, val) in params
+        if val isa Dict &&  ~(key in ["weights", "e0"])
+            params[key] = fill_defaults!(val; param_key = key)
+        elseif  key == "basis"
```
gelzinyte: Maybe it would be better to organise this as 

```
fit_params["basis"] = Dict(
    "pair_basis" => pair_basis_param_dict,
    "rpi_bsasis" => rpi_basis_param_dict) 
```
instead of the following?
```
fit_params["basis"] = [
    "pair_basis" => pair_basis_param_dict,
    "rpi_bsasis" => rpi_basis_param_dict]
```

The former would make it easier to fill in, the latter is a bit more general.

gelzinyte:
Thank you very much for all the feedback! I think I have followed up to all of your suggestions - looking forward to any other comments.

scripts/generate_artifacts.jl
cortner: shall I go ahead and merge ACEData#1?

src/read_params.jl
cortner: Actually it is not really more general, we can always loop 
```
for (basisname, basisdict) in fit_params["basis"]
    ... 
end
```

cortner: The problem with the second approach is that it is not "raw" JSON. I would definitely go with the first.

scripts/generate_artifacts.jl
gelzinyte: actually, let me first update the yaml file and structure of basis_params

cortner:
@gelzinyte you may want to put a lower version bound  0.11.4 for JuLIP so that the YAML works.

src/read_params.jl
gelzinyte: Of course, it's also trivial to iterate over a dictionary.. Updated.

scripts/generate_artifacts.jl
gelzinyte: ACEData#1 is up to date with this PR. However, it is probably easier to hold on from merging it until this is ready to be merged, in case addressing any other comments would require changing some of the files.

gelzinyte:
@cortner do you have any further feedback? Maybe it would be good to merge and open this up to others and add any further changes as separate pull requests?

src/fit.jl
```diff
@@ -31,6 +31,12 @@ function fit_ace(params::Dict)
         solver["P"] = ACE1pack.generate_precon(basis, params["P"])
     end
 
+    if typeof(params["e0"]) == Dict{Any, Any}
+        # sometimes gets read in (from yaml?) as Dict{Any, Any} 
+        # which gives StackOverflowError somewhere in OneBody
+        params["e0"] = convert(Dict{String, Any}, params["e0"])
```
cortner, line 37: sounds like it should be filed as an issue with JuLIP

cortner:
@gelzinyte all looks good.  Just need to confirm a licensing issue with @gabor1  and then we should be ready to merge. Thank you for taking the lead on this. I'll wait for your instruction though to first merge the data repo and then this here. We can try to wrap it up tomorrow (Tuesday).

cortner:
hm .... I'm struggling with the Artifacts...
```
┌ Warning: using Pkg instead of using LazyArtifacts is deprecated
│   caller = ip:0x0
└ @ Core :-1
  Downloaded artifact: TiAl_tiny_dataset
 Downloading artifact: TiAl_tiny_dataset
┌ Error: Hash Mismatch!
│   Expected sha256:   42f96e718e7b8d1f3e9bbbe0b7c2be1b3bfe96c645373af5ac6eeebc997629e2
│   Calculated sha256: 5c5ddd4883d50d1f5ce8c6bacc36ec15ffd08d96be0b774fea7262db14e1ba25
└ @ Pkg.PlatformEngines ~/gits/julia17/usr/share/julia/stdlib/v1.7/Pkg/src/Plat
  Downloaded artifact: TiAl_tiny_dataset
ERROR: LoadError: Unable to automatically install 'TiAl_tiny_dataset' from '/Users/ortner/gits/ACE1pack/Artifacts.toml'
Stacktrace:
```

cortner:
and even if I fix the sha and switch to LazyArtifacts, I get this: 
```
  Downloaded artifact: TiAl_tiny_dataset
  Downloaded artifact: TiAl_tiny_dataset
ERROR: LoadError: Unable to automatically install 'TiAl_tiny_dataset' from '/Users/ortner/gits/ACE1pack/Artifacts.toml'
Stacktrace:
...
```

cortner:
thoughts?

gelzinyte:
I think the problem was that the line

```
tarball_hash = archive_artifact(hash_, joinpath(tarfile))
```
was modifying the `tarfile` and as a result, `tarball_hash` was different from the hash for the file at the url. But `(url, tarball_hash)` were bundled together. Which is why hash for file at downloaded didn't match the saved key. I didn't have this problem initially, because the files on my computer were downloaded with `generate_artifacts.jl` and weren't checked against not-yet-defined incorrect hash.

The problem appeared only after removing the appropriate directories under `~/.julia/artifacts` and re-runing `artifacts.jl` to trigger a new download. I followed the ACE `create_artifacts.jl`, so that repo might have the sane problem? 

My workaround was to generate the hash myself, although I'm surprised that I had to, so maybe something's not as it's supposed to be? Also, a (small?) caveat - this relies on any downloaded artefacts being already compressed or small. 


Unfortunately, something's still wrong and the error message isn't very helpful: 
```
(main) ➜  test git:(eg/dev) ✗ julia artifacts.jl 
  Downloaded artifact: TiAl_tiny_dataset
  Downloaded artifact: TiAl_tiny_dataset
ERROR: LoadError: Unable to automatically install 'TiAl_tiny_dataset' from '/Users/elena/.julia/dev/ACE1pack/Artifacts.toml'
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] ensure_artifact_installed(name::String, meta::Dict{String, Any}, artifacts_toml::String; platform::Base.BinaryPlatforms.Platform, verbose::Bool, quiet_download::Bool, io::Base.TTY)
   @ Pkg.Artifacts /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/Pkg/src/Artifacts.jl:441
 [3] ensure_artifact_installed(name::String, artifacts_toml::String; platform::Base.BinaryPlatforms.Platform, pkg_uuid::Nothing, verbose::Bool, quiet_download::Bool, io::Base.TTY)
   @ Pkg.Artifacts /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/Pkg/src/Artifacts.jl:404
 [4] _artifact_str(__module__::Module, artifacts_toml::String, name::SubString{String}, path_tail::String, artifact_dict::Dict{String, Any}, hash::Base.SHA1, platform::Base.BinaryPlatforms.Platform, lazyartifacts::Any)
   @ Artifacts /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/Artifacts/src/Artifacts.jl:547
 [5] invokelatest(::Any, ::Any, ::Vararg{Any}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Base ./essentials.jl:716
 [6] invokelatest(::Any, ::Any, ::Vararg{Any})
   @ Base ./essentials.jl:714
 [7] top-level scope
   @ /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/Artifacts/src/Artifacts.jl:680
in expression starting at /Users/elena/.julia/dev/ACE1pack/test/artifacts.jl:4
```

cortner:
so we are still stuck here? ...

cortner:
I thought the hash can be just anything (compatible with whatever keys they use...) but you are saying this is not the case?

scripts/generate_artifacts.jl
```diff
@@ -0,0 +1,66 @@
+
+using ACE1pack, Pkg
+using Pkg.Artifacts
+using SHA
+
+artifacts_toml = joinpath(pathof(ACE1pack)[1:end-16], "Artifacts.toml")
+
+function _get_sha256(filename)
```
gelzinyte, line 8: Is this sound? Will `return` close the file? 

Alternatively, I thought about 
```
function _get_sha256(filename)
    open(filename) do f
        global hash = bytes2hex(sha256(f))
    end
    return hash
end
```
but it seemed wrong to make `hash` accessible across the script. 

It wasn't clear to me what is the correct way to access anything derived from the file contents outside of the `do` block. Maybe in that case one should explicitly `close()` the file?

cortner: I'm afraid I have no memory about how this works. Have you checked what I did in `ACE1.jl`?

cortner: Artifacts.toml: 
```
["acedata_v0.8.1"]
git-tree-sha1 = "23a2e3400188336105a4e47c31cdb7ce95a104cb"
lazy = true

    [["acedata_v0.8.1".download]]
    sha256 = "91305df8a0b0572630b9be4b2f3b7e16fc14b2683182e25f18a338036eac0305"
    url = "https://github.com/ACEsuit/ACEData/archive/v0.8.1.tar.gz"
```

and then in `artifacts.jl`: 
```Julia
using Pkg.Artifacts

datadir = joinpath(artifact"acedata_v0.8.1", "ACEData-0.8.1")
testsdir = joinpath(datadir, "tests")
traindir = joinpath(datadir, "trainingsets")
potsdir = joinpath(datadir, "potentials")

# unzip the eam potential
if !isfile(joinpath(traindir, "w_eam4.fs"))
   eamzip = joinpath(traindir, "w_eam4.fs.zip")
   run(`unzip $eamzip -d $traindir`)
end
```

cortner: why do you even need the sha in `artifacts.jl`?

gelzinyte: > why do you even need the sha in artifacts.jl?

`bind_artifact!()` needs both sha and url and you need the url to get the artefact. 

> I thought the hash can be just anything (compatible with whatever keys they use...) but you are saying this is not the case?

hashes for the downloaded file and file used to create the artefact need to match, no? I thought the initial error was raised because they didn't. 

ACE1 artefacts are successfully downloaded with what I thought was equivalent code, so maybe the issue is with how the original artefact files were set up. I'll see if I can make any progress tomorrow

gelzinyte:
The issues with artefacts only relate to tests. Since there is another PR coming to fix the GitHub's CI, what do you think about fixing the artefacts there as well?

cortner:
ok, then I'll merge and tag a first version. We can go back to all remaining issues afterwards.

cortner:
thank you again for all your work on this.

cortner:
done, this is tagged as 0.0.1

PR #14, opened 2022-03-09
Generate artefacts

gelzinyte:
The `scripts/generate_artifacts.jl` now updates `Artifacts.toml` based on the tarball hosted at the given (GitHub) link. Whenever the artefact is used in scripts, the same file gets downloaded and decompressed. 

I think the artefact-related problems we had in PR #1 are resolved, but @cortner would you mind running the tests locally to check? 

If all looks good, [ACEData PR #1](https://github.com/ACEsuit/ACEData/pull/1) can be merged. Then I'll update the links here and merge this PR.

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -4,7 +4,7 @@ lazy = true
 
     [[ACE1pack_test_files.download]]
     sha256 = "5416b14aa2356ce8a399102a45232e8c44e746c3270c4baa15d4bbcce9fcc2e7"
-    url = "https://github.com/gelzinyte/test_julia_artifacts/blob/main/ACE1pack_test_files.tar.gz?raw=true"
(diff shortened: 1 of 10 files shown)


cortner:
Some notes ... 
- when you do regression on a different hardware, with different blas, then the errors won't be the same to within machine precision, so you can only test error <= 2 * expected_error or something like that. I'm trying this now.
- I usually like to see when a test passes to I print a green + or Test Passed, that's what print_tf and println_slim are for. But this is just a personal preference, you definitely don't need to follow it, just to explain some of the edits I'm making.
- I make sure that the test scripts run both inside a @testset as well as independently, makes debugging much easier. That's another set of edits I'm making. 


I'll push this in a moment.

cortner:
I had to make a few small changes, but the tests now pass locally

gelzinyte:
Thank you very much for the tips! I have updated the download links and I think everything works as it should, so I will merge this in a bit.

Issue #15, opened 2022-03-11
Flexible degrees


cortner:
interface to per order degrees + related

gelzinyte:
Resolved in #28

PR #17, opened 2022-03-15
specify polynomial degree for each correlation order and species

gelzinyte:
Added a `sparseM` polynomial degree measure that allows to specify a different polynomial degree for each correlation order and species. Issue #15.

The default is `type = "sparse"` in `degree_params` and the polynomial degree is specified with `maxdeg` in `basis_params`. So to get "sparse" degree parameters: `degree = degree_params()`

`degree_params` with `type = "sparseM"` has to have `Dd` specified (for example at least `Dd = Dict("default" => 17)`), which is how the polynomial degree is specified, so `maxdeg` will in most instances be set to `1.0`.  Then `degree = degree_params(type = "sparseM", Dd = Dict("default" => 17))`  with `maxdeg = 1.0` is equivalent to the "sparse" case above.

A more extended example: 

```
zH = AtomicNumber(:H)
zC = AtomicNumber(:C)
Dn = Dict( "default" => 1.0 )
Dl = Dict( "default" => 1.5 )
Dd = Dict( "default" => 17,
        1 => 17,
        2 => 15,
        (3, zC) => 12, 
        (3, zH) => 4) 

degree = degree_params(
    type = "sparseM",
    Dd = Dd, 
    Dl = Dl, 
    Dn = Dn)
```

So that we pick an intuitive name - what does the "M" stand for in "SparsePSHDegreeM"?


gelzinyte:
Just a note - forgot to convert atomic numbers to and from strings, will update

cortner:
do you convert atomic numbers to string containing the chemical symbol? That seems the right approach to me. Though could also just have a `read_dict, write_dict` for `AtomicNumber` that does the conversion automatically.

gelzinyte:
I'll

gelzinyte:
Yes I agree, I'll convert atomic numbers to string containing the chemical symbol. That's also what we do to species involved in the rest of making the basis.

gelzinyte:
@cortner could you please take a look?

cortner:
@gelzinyte ouch, this has been lying around for a while. Is this ready to review and merge?

src/read_params.jl
```diff
@@ -8,6 +8,9 @@ export  fill_defaults!
 
 """ Recursively updates nested dictionaries with default parameters"""
 function fill_defaults!(params::Dict; param_key = "fit_params")
+    # THIS IS THE LINE FROM THE polydeg-per-N BRANCH - HOW SHOULD THIS BE
```
cortner: @gelzinyte --- I had to resolve a merge conflict here, but didn't know how to correctly merge. Can you please take a look at this?

gelzinyte: everything from both branches should have been kept - I've pushed the correct list

PR #24, opened 2022-04-26
Add tests for more solvers and tighten test tolerances

wcwitt:
@gelzinyte, I'm hoping to tweak and tighten the tolerances for this test. Sound okay?


gelzinyte:
Certainly, the previous form wasn't quite right, I think.

wcwitt:
Great, thanks. Any guesses as to why `atol = 1e-3` is the lowest I can go right now? Even if I restrict it to the RMSEs and skip the relative errors.

gelzinyte:
The values compared against are what I had when fitting on my laptop and running tests locally still gives the same results within machine precision. So maybe the differences are because of different architectures and related libraries? I don't have any sense of how much the accuracy is expected to change with different set-ups, but `atol = 1e-3` seems high?

wcwitt:
Thanks, that's helpful. I'd be more comfortable with `atol=1e-6` for the RMSEs. I'll try on other machines too. Maybe we'll need to change the LSQR tolerances.

cortner:
I don't have a view on this so happy for you to merge whenever ready.

cortner:
just to document this here as well: after today's ACEfit meeting we decided to check whether the tests become reproducible if we enforce IEEE arithmetic.

wcwitt:
@gelzinyte @cortner, I propose we merge this now. I added new tests (for a tiny silicon dataset) that cover the QR, LSQR, RRQR, BRR, and ARD solvers. I've been using these tests with ACEfit, so adding them here is a step towards sunsetting IPFitting.

I've set strict tolerances (atol=1e-5 for config-level RMSEs) for all solvers except LSQR. For reasons I still don't understand, LSQR has more machine-dependence than the others. I've tried `--math-mode=ieee`, which does change the RMSEs slightly, but doesn't eliminate the machine dependence. For now, I've added a warning saying the LSQR test tolerance is loose - maybe we'll figure it out at some later point.

wcwitt:
Blast, after saying that I now see the ARD test failed on CI.

PR #21, opened 2022-04-20
WIP: ACEfit integration

cortner:
CC @wcwitt, @gelzinyte  - I'm going to get this started and put together a first 1-2 examples that use both and then we can discuss here how to move everything over.

PR #31, opened 2022-05-27
Literate Tutorials

cortner:
here I tried to generate the tutorials automatically via Literate.jl. Some notes: 
* The literate tutorials should just be stored as scripts, then used to generate the md files and those md files should NOT be checked into git !!
* I also rearranged the documentation a little bit and removed some pages that should likely be compressed into a single one or into tutorials.
* I've removed all ACEfit.jl from this branch, it will come back in with #21 later.

TODO
- [x] The TiAl tutorial doesn't execute correctly, I think I tis because it can't find the artifact. Apparently I still don't understand artifacts - where does it looks for the `Artifacts.toml`???
- [ ] I've kept the original versions of the first two tutorials in the repo in case they have been edited / improved since I branched off from main. But those should be removed before merging.


cortner:
the last commit adds a `artifact(str)` function to `ACE1pack` which then just calls the `@artifact_str` macro. The point is that from within `ACE1pack` the correct `Artifacts.toml` is accessible so now we can load those example training sets from anywhere.  (in particular from the docs)

cortner:
To merge this I would be grateful, @gelzinyte, if are you able to take a quick look at the tutorials and make sure I didn't miss any updates?

cortner:
good idea with the ignored folder

gelzinyte:
also cleaned up the section headers - they only appear correctly in the sidebar if they start with `#`, `##`, etc. 

Otherwise looks good, thanks! I really like literate tutorials - they just ran for me without any problem.

gelzinyte:
I will start adding ACE1pack documentation once this is merged

cortner:
we can merge now if you wish?

PR #39, opened 2022-07-25
added ace_fit script

WillBaldwin0:
as above

### DIFF
--- a/scripts/ace_fit.jl
+++ b/scripts/ace_fit.jl
@@ -1,0 +1,53 @@
+using Pkg; Pkg.activate(".")
+
+using ACE1pack, ArgParse
+
+parser = ArgParseSettings(description="Fit an ACE potential from parameters file")
+@add_arg_table parser begin
+    "--fit-params", "-p"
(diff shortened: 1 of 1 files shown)


scripts/ace_fit.jl
```diff
@@ -0,0 +1,53 @@
...
+    num_observations = get_num_observations(fit_params["data"])
+    basis_size = get_basis_size(fit_params["basis"])
+    dry_fit_filename = replace(fit_params["ACE_fname"], ".json" => ".size")
+    size_info = Dict("lsq_matrix_shape" => (num_observations, basis_size))
+    save_json(dry_fit_filename, size_info)
+    exit(0)
+end
+
+if haskey(ENV, "WFL_ACE_FIT_BLAS_THREADS")
+    using LinearAlgebra
```
gelzinyte, line 38: This is workflow specific. Should we keep it with a rename `WFL_ACE_FIT_BLAS_THREADS`> `ACE_FIT_BLAS_THREADS`?

Issue #43, opened 2022-07-26
ArgParse


gelzinyte:
`ace_fit.jl` uses `ArgParse` which needs to be `add`ed by the users. What would be the best way to just have it as ACE1pack's dependency and not need for the users to install it alongside ACE1pack? 

I would have ArgParse be re-exported by ACE1pack and in the script call `using ACE1pack.ArgParse`. Any better ideas? @WillBaldwin0, @cortner, @wcwitt

cortner:
why does it have to be added by the users? To run the cmd line  script?

cortner:
If that's the case then your suggestion is the right thing to do I think.

cortner:
I'm happy to tag a 0.0.4 if you want to fix this now quickly.

gelzinyte:
> why does it have to be added by the users? To run the cmd line script?

Yes, exactly. 

Thanks, I'll open a PR shortly

wcwitt:
This looks resolved to me, thanks @gelzinyte.

PR #44, opened 2022-07-26
updated argparse to work behind the scenes

gelzinyte:
Fix for #43. @cortner please merge if this looks ok and update the versions. Seems to work in a fresh environment for me.

### DIFF
--- a/scripts/ace_fit.jl
+++ b/scripts/ace_fit.jl
@@ -1,4 +1,5 @@
-using ACE1pack, ArgParse
+using ACE1pack
+using ACE1pack.ArgParse
 
 parser = ArgParseSettings(description="Fit an ACE potential from parameters file")
 @add_arg_table parser begin
@@ -36,6 +37,7 @@ function save_dry_run_info(fit_params)
     exit(0)
 end
 
+args = parse_args(parser)
 nprocs = args["num-blas-threads"]
 if nprocs > 1
     using LinearAlgebra
@@ -43,7 +45,6 @@ if nprocs > 1
     BLAS.set_num_threads(nprocs)
 end
 
-args = parse_args(parser)
 raw_params = load_dict(args["params"])
 fit_params = fill_defaults(raw_params)
 
--- a/scripts/ace_fit.jl
+++ b/scripts/ace_fit.jl
@@ -1,5 +1,4 @@
 using ACE1pack
-using ACE1pack.ArgParse
 
 parser = ArgParseSettings(description="Fit an ACE potential from parameters file")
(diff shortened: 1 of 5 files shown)


cortner:
looks good, but maybe can be simplified. It's late so I'm tempted to just merge, but I think the fact we reexport everything means you don't even need to do the `using ArgParse`? Have you tried this?

gelzinyte:
Let me quickly try that

cortner:
Do you want to very quickly hop onto Zoom?

cortner:
@WillBaldwin0 -- we can merge and tag this PR now or wait until your workshop is over. Let us know.

WillBaldwin0:
So will this also need pyjulip to be updated because pyjulip current does using ACE1 etc?

cortner:
thanks for flagging this. Let's not merge then and figure out how to handle this. 

CC @casv2

cortner:
my suggestion is that pyjulip could just go through a list of packages, check which are available, and import those. And if something doesn't work, throw an error message saying "most likely one of the required packages are not installed" ... or something along those lines?

gelzinyte:
If I understand it correctly, this merge and a bump in version (which would also fix a bug) is waiting for pyjulip to be updated?

casv2:
Just merge this we can make pyjulip work with however this is decided to work.

We'll have to look for the packages and raise an error if it's not there? Actually, pyjulip looks for ACE1 and JuLIP, but those are still around in ACE1pack right?

gelzinyte:
Yes, ACE1pack re-exports ACE1, IPFitting and JuLIP. So now `using ACE1pack` is sufficient if you wanted to use any of the ACE1, JuLIP, etc functions. e.g. 

```
using ACE1pack 
db = LsqDB(...)
```
is enough and you don't have to `using IPFitting`. (in fact, that will raise an error, because formally IPFitting wasn't `added`. 

I guess pyjulip has to work with either a) only ACE1pack installed and b) ACE1, JuLIP, IPfitting, but not ACE1pack installed. 

If pyjulip is being updated v soon I can update the docs and merge this if @WillBaldwin0 and @cortner are happy?

The docs update: change one-liner to only install ACE1pack. Anything else?

casv2:
If you're changing the docs to just install ACE1pack shouldn't pyjulip also just go through ACE1pack and go from there? 

Rename to pyACE1pack?

gelzinyte:
once this is merged, @cortner could you also bump the version? (to 0.0.4)

cortner:
done

PR #50, opened 2022-08-16
Fix typos

WillBaldwin0:
@cortner @gelzinyte, I've made some small changes to a lot of files, fixing some typos and making sure everything reads and links together correctly. Also added the python usage. 

I noticed that for those only using command line, we'll need to add a script to ACE1pack which just calls the export function to convert a `.json` to a `.yace`. I'm happy to write that, or we could make `ace_fit` automatically export to `.yace`.


wcwitt:
Thanks a lot Will.

> or we could make ace_fit automatically export to .yace

That sounds good to me.

cortner:
so it will produce two files, a json and a yace? I'm defintely ok with that.

WillBaldwin0:
I will merge this if no one else has any comments?

WillBaldwin0:
Also, since I've modified the ace_fit script to do the automatic exporting, do we need to take a version?

gelzinyte:
Looks good, thanks a lot! I think exporting `yace` automatically is more convenient.

PR #51, opened 2022-08-17
Fix typos

WillBaldwin0:
after merging, the docs were not building due to a bug which I didn't see in the tests on the PR. 

bug was to do with the fact that in ACE1pack the orebody, pair and ace can be in an arbitrary order. The export function now handles this.


cortner:
this looks great, thank you.

cortner:
let's not do it now, but as an extension you could consider what to do if there is not one-body or two-body?

PR #64, opened 2022-09-29
Test for ace_fit.jl

wcwitt:
First attempt at #60.

@gelzinyte & @bernstei should we expand this at all? It was the quickest thing I could do to start.

Note it applies to the IPFitting version, hence the PR to `0.0.x`. Once we agree on the test, I'll adapt it for the ACEfit version.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -16,4 +16,6 @@ using ACE1pack, Test, LazyArtifacts
 
     @testset "Test silicon" begin include("test_silicon.jl") end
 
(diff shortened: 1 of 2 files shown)


bernstei:
Looks fine to me. Is there a full-features testing framework (analogous to `pytest`) that this should possibly be converted to?  Right now it has things like bare `julia` executable and relative paths that might not work in such a framework.

wcwitt:
I think the test framework we have (e.g., with @testset and @test macros) is the usual Julia one. But not confident what I've done with `run(pipeline(...` is optimal.

PR #72, opened 2022-10-11
ace_fit.jl control numpy threads 

### LINKED ISSUE #71: set number of blas threads more widely
https://github.com/ACEsuit/ACE1pack.jl/blob/a51295085a9fe45ea588d8cfb81a9fcb3f009e1a/scripts/ace_fit.jl#L48

Would it be helpful, or just annoyingly complicated, if `ace_fit.jl` set all the known BLAS env vars for multithreading, i.e. `MKL_NUM_THREADS`, `OPENBLAS_NUM_THREADS`, maybe also plain `OMP_NUM_THREADS`, since some of the optimizers (namely the ones in `scikit-learn`) use those instead of the julia-internal value?

closed by #72

bernstei:
ace_fit.jl set all possible env vars controlling numpy numbers of threads when --num-blas-threads is passed

closes #71

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,6 +20,7 @@ MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
 MPIClusterManagers = "e7922434-ae4b-11e9-05c5-9780451d2c66"
 PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
 ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
+PyCall = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Suppressor = "fd094767-a336-5f1f-9728-57cf17d0bbfb"
@@ -29,8 +30,8 @@ YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
 [compat]
 ACE1 = "= 0.9.2"
 JuLIP = "0.11.4"
-julia = "1.6"
 PrettyTables = "1.3"
+julia = "1.6"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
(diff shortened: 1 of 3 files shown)


bernstei:
Could we actually just call `threadpool_info` mentioned here https://stackoverflow.com/questions/30791550/limit-number-of-threads-in-numpy?

wcwitt:
Have you managed to verify whether setting the environment variables this way works? I assume yes, but if they need to be set before `import numpy` (perhaps only in older python?), I'm not exactly sure what that means for us.

bernstei:
I haven't.  Note that `threadpool_limits` is definitely supposed to work after import.  Let me do some tests.  If `threadpool_limits` works with MKL, do we consider that sufficient to just go with that?

wcwitt:
The `threadpoolctl` option looks nice too - maybe better because we could print the `threadpool_info()` just before calling the solver to verify. We'd need to figure out how to do `with threadpool_limits(limits=1, user_api='blas'):` in Julia, or something equivalent.

Edit: I wrote this before seeing your last comment. But this sounds good to me:

> If threadpool_limits works with MKL, do we consider that sufficient to just go with that?

bernstei:
I guess it's a python context manager, so it's not at all obvious how to call it from julia, which I gather is using `pyimport` to call particular python functions (specifically the sklearn optimizer constructors, which then create julia objects with `fit()` methods).

bernstei:
I just noticed that the version of `ace_fit.jl` that my julia is actually using and this repo version are very different.  Do you know why that is?

wcwitt:
> I just noticed that the version of ace_fit.jl that my julia is actually using and this repo version are very different. Do you know why that is?

Yours should match the 0.0.x branch. But I'll make sure these changes go both places.

wcwitt:
For now, give me a day or two to do these?

https://github.com/ACEsuit/IPFitting.jl/issues/39
https://github.com/ACEsuit/ACEfit.jl/issues/35

That will help us know when we've solved the problem

bernstei:
Definitely - if you can get `threadpoolctl` to work, that'd be much cleaner.

wcwitt:
I think this is working now. Here's the version for IPFitting: https://github.com/ACEsuit/ACE1pack.jl/pull/74.

Edit: looks like the CI is failing here, but for an unrelated reason. The tests in the more stable #74 passed.

bernstei:
Great, thanks @wcwitt

bernstei:
Do you want me to test it? If so, can you tell me how to switch to the relevant version in my julia configuration?

Issue #84, opened 2022-12-16
ACE registry


cortner:
... is now available at 
```
https://github.com/ACEsuit/ACEregistry
```
This should now be used instead of the `MolSim` registry. Can somebody please update the docs?

tjjarvinen:
I will take care of this.

PR #85, opened 2022-12-17
Online documentation update

### LINKED ISSUE #83: Current Stable Documentation is not Working.
The documentation section that points to stable version is not working. Because it point to an ancient version. This is an issue because it is the version that is linked from the main GitHub page. For new people starting with ACE this is big issue, as it can cause a lot of confusion with non working documentation.

There are two possible fixes

1. Remove the links to the stable version and have all links point to the developer version.
2. Tag version to GitHub, which will trigger stable version build for the documentation.

Option 2. is probably the best way, as it does not need anything else than a tag in the repo.

tjjarvinen:
This will fix #83 and changes the documentation to point to the new registry (#84).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -52,7 +52,7 @@ jobs:
             ${{ runner.os }}-
       - run: |
           using Pkg
-          Pkg.pkg"registry add https://github.com/JuliaMolSim/MolSim.git"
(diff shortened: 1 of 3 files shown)


cortner:
maybe also worth explicitly commenting on the fact that this is still under development and only the `dev` version of the documentation is up to date.

PR #89, opened 2023-01-10
WIP: Polishing the Docs

cortner:
... based on feedback from tutorials at UBC.


tjjarvinen:
Should the installation instruction to be more concise? Something like

```julia
using Pkg
pkg"activate ."
pkg"registry add https://github.com/ACEsuit/ACEregistry"
pkg"add ACE1pack"
```

This would make it easy to copy-paste.

cortner:
maybe a good idea. Make the change and push to this PR?

PR #92, opened 2023-01-16
WIP: New Affine Model Interface 

cortner:
This implements the proposed affine model interface. @tjjarvinen and I think this will make our models far more accessible from Julia and maybe this can also guide simplification of the python/json interface. 

Here is what a minimal script now looks like: 
```julia 
using ACE1pack
data = read_extxyz(joinpath(ACE1pack.artifact("TiAl_tutorial"), "TiAl_tutorial.xyz"))
model = acemodel(species = [:Ti, :Al], N = 3, maxdeg = 15, rcut = 5.5, 
                 maxdeg2 = 10, Eref = Dict(:Ti => -1586.0195, :Al => -105.5954))
acefit!(model, data[1:5:end], ACEfit.RRQR(rtol = 1e-4))
# export_ace("TiAl.yace", model)
```

My goal is that, by the time the software paper is published, these few lines give a very good, robust model with repulsive core. 

The last line is commented because this doesn't work yet. @casv2 and @WillBaldwin0  -- can you please coordinate providing this functionality? I suggest that we merge this first and you can do it in a separate PR.


cortner:
@wcwitt  -- should we use a different solver as the default?

cortner:
@wcwitt  --- sorry to ping you again. Can you look at the minimal script above and suggest what the default solver should be?

wcwitt:
Sorry - will do a full review as soon as possible. For the solvers, is there a reason you chose RRQR? ACEfit has a few Bayesian solvers, if that's what you actually want, and I think BLR could be reasonable as a default also.

I think I need to rename some things - right now I have slightly different names for BRR-variants using different factorization strategies (e.g. SVD or Cholesky), and things aren't as clear as they should be

cortner:
BLR or ARD and any parameters?

wcwitt:
I would say BLR (it's notably faster than ARD) with default parameters.

But linking this issue https://github.com/ACEsuit/ACEfit.jl/issues/41.

gelzinyte:
I don't think there will be much of a change in the user-facing side of ACE1pack-julia/json interface, because all of the same parameters still need to be specified (but let me know if I miss something here). But how ACE1pack parameters are translated into `acemodel` will need to be updated, I'll wait for this to be merged first.

cortner:
@gelzinyte  --- I'm not even so sure that this new model interface will change much (or anything?) for the JSON interface. What do you think?

gelzinyte:
Yes, I agree. I was talking about switching to using `acemodel` in `ACE1pack.fit_ace`, for example. I don't think that will change too much either.

cortner:
add alternative tutorials that use this new interface, then merge it.

PR #100, opened 2023-01-24
Add adding General registry in the short installation instructions

gelzinyte:
I think adding General registry was missed out from the short installation instructions. @cortner please merge this PR if this was actually unintentional.


cortner:
@tjjarvinen  -- I think we discussed this. Can you please comment why we left it out? 

I believe the context of Elena's request is that for first-time Julia users the `General` registry doesn't get installed if the very first command is to add the ACE registry.

gelzinyte:
> I believe the context of Elena's request is that for first-time Julia users the General registry doesn't get installed if the very first command is to add the ACE registry.

Indeed, see output:

```
(ace_project) pkg>  registry add https://github.com/ACEsuit/ACEregistry
     Cloning registry from "https://github.com/ACEsuit/ACEregistry"
       Added registry `ACE` to `~/.julia/registries/ACE`

(ace_project) pkg> add ACE1pack
    Updating registry at `~/.julia/registries/ACE`
    Updating git-repo `https://github.com/ACEsuit/ACEregistry`
   Resolving package versions...
ERROR: expected package `Calculus [49dc2e85]` to be registered

(ace_project) pkg> registry add https://github.com/JuliaRegistries/General
     Cloning registry from "https://github.com/JuliaRegistries/General"
       Added registry `General` to `~/.julia/registries/General`

(ace_project) pkg> add ACE1pack
   Resolving package versions...
   Installed Calculus ───────────────────── v0.5.1
   Installed x265_jll ───────────────────── v3.5.0+0
...
```

cortner:
I guess there is no harm in adding that line. But please add a comment below that this step is only needed when using Julia for the first time so that nobody gets confused.

cortner:
@gelzinyte - do you mind adding a comment below or above the revised installation instructions that adding  `General` is only needed if this install is undertaken at the very first startup of julia?

wcwitt:
On this topic (roughly), is it correct that the `pkg"activate ."` syntax is deprecated in favor of `Pkg.activate`? I get warnings about this sometimes

And similar for `pkg"registry add https://github.com/JuliaRegistries/General"`

tjjarvinen:
> @tjjarvinen -- I think we discussed this. Can you please comment why we left it out?
> 
> I believe the context of Elena's request is that for first-time Julia users the `General` registry doesn't get installed if the very first command is to add the ACE registry.

I think our discussion was that it was an old bug that you need to add General Registry and it is fixed by now. But it might be present for the first install.



> On this topic (roughly), is it correct that the `pkg"activate ."` syntax is deprecated in favor of `Pkg.activate`? I get warnings about this sometimes
> 
> And similar for `pkg"registry add https://github.com/JuliaRegistries/General"`

I have not heard anything about this and I have not got warnings when using it. But I see no problem going to `pkg.activate(".")` style.

edit. Now that I remember it. You should only use `pkg"activate"` when using console, if used in code it can cause problems.

wcwitt:
I remembered one place I'd seen the warnings ... in the CI

image

wcwitt:
This is simple enough, and the non-MPI tests are passing, so I'm going to merge.

PR #104, opened 2023-01-30
WIP: Backporting simplified multi-degree interface

cortner:
this backports the simplified multi-degree interface that I put into ACE1 0.11.x (and therefore ACE1pack 0.3.x). Maybe I'll add another test or two but then this can be merged. The branch dev-0.2.x can be used to backport any other important functionality (if any)


cortner:
@wcwitt  ---  I'd like to merge this (and tag as the next 0.2.x) if you confirm that this doesn't interfere at all with how Gabor's group uses ACE1pack?

wcwitt:
Sorry to be slow with this - what do you propose doing now since we've moved beyond 0.2.x?

cortner:
close since no longer needed?

PR #121, opened 2023-04-24
[WIP] LAMMPS export

wcwitt:
Not many commits yet, but I started mapping this out today and think I know what is required. Here's my PACE fork: https://github.com/wcwitt/lammps-user-pace. I'll try to finish over the next few days.

Some lingering questions:

* @cortner, I need to extract `c0 + c1*(r-ri) + c2*(r-ri)^2 + c3*(r-ri)^3` coefficients for each bin of a spline. Do I need (roughly) the ForwardDiff syntax from `ACE1.Splines.export_splines` or is there a better way? Edit: I meant `ACE1.Splines.evaluate_d`.

* Should we transition PACE to completely to splines for our potentials, or try to maintain backwards compatibility?

* Is there a specification for the PACE `yaml` format anywhere?

* I haven't found any `export_ace` routines that work with `SumIP` potentials. Have I missed anything, and if not what should be the strategy?


cortner:
1. I will help you with that.
2. yes - transition now and never look back
3. I have no idea, please email Yuri
4. That's odd. Cas should know how to deal with that though.

wcwitt:
1. I know what to do now - no worries. You can check it before we merge.

casv2:
There's more information here, but currently we take care of the pair potential separately. I don't think we really have to do anything here.

https://github.com/ACEsuit/ACE1pack.jl/blob/72821f403511b064ed3233f96113ea68624a6e81/src/export_multispecies.jl

Here's an example (the overlay)

https://acesuit.github.io/ACE1pack.jl/dev/tutorials/lammps/

wcwitt:
Todo before merging: add the `export_lammps` option to `acefit!` and related standalone function.

wcwitt:
Reminder to add the spline resolution as an option to `export2lammps`

src/export.jl
```diff
@@ -1,231 +1,323 @@
...
+
+        if typeof(V2.basis.J) <: SMatrix
+            get_ru(jj) = jj.ru
+            rus = get_ru.(V2.basis.J)
+            rout = maximum(rus)    
+        else
+            rout = V2.basis.J.ru
+        end
+
+        rin = 0.1
```
casv2: I propose reducing this to 0.001 or perhaps even smaller. We need it in NS. 

Perhaps it's also good to replace the pair export with non-linear splines too at some point too...

casv2:
Thanks Chuck, I can confirm this works well for me now and can be merged

src/export.jl
wcwitt: Done. We can revisit if it remains a problem. Is this change (plus the bugfix) sufficient for the nested sampling to succeed reliably?

I'd be happy with the non-linear splines, but slightly happier with a second `yace` that consolidates all the two-body terms. I will ping Yury again to see if he thinks that is feasible.

wcwitt:
There are a few things that still need addressing before this is finished. The `model.jl` page has some TODOs, for instance.

cortner:
do you need my input to wrap this up?

wcwitt:
Let's merge this for the sake of moving forward and address any remaining issues as they come. @cortner would you mind tagging a new version?

Issue #132, opened 2023-06-22
PythonCall update in ACEfit


wcwitt:
The tests are failing now because of the switch to PythonCall in ACEfit. I've tried to fix `test_silicon.jl` (for example), but I keep getting `PythonCall`-related segfaults. Any ideas @tjjarvinen?

tjjarvinen:
To add a TODO here.

- [x] ASE.jl needs to be either removed or changed to use PythonCall
- [x] Remove PyCall from Project.toml
- [x] Remove pip usage from CI - if we use PythonCall we should use [CondaPkg.jl](https://github.com/cjdoris/CondaPkg.jl)
- [x] Add Project.toml to tests - this is Julia recommendation that all packages should use.

Might be something more too, but this is a start

wcwitt:
I believe #134 addressed everything here except the `pip` usage in the CI. But I don't have time to look at that just now.

cortner:
@wcwitt  -- can we try to remove ASE.jl? Is anybody still using this?

wcwitt:
I removed `ASE.jl` as part of #134. It seemed to improve #131, which I was glad to see. Hopefully the removal doesn't affect anyone.

cortner:
oh great, this is already tagged as 0.4.1. And testing locally seemed fine.

cortner:
can this be closed?

wcwitt:
@tjjarvinen suggested this, which is fine with me, although I don't know how to do it quickly myself. Everything else is done I think.

> Remove pip usage from CI - if we use PythonCall we should use [CondaPkg.jl](https://github.com/cjdoris/CondaPkg.jl)

tjjarvinen:
Now that ASE has been removed and ACEfit uses CondaPkg.jl for sklearn, we can just delete the pip line from CI.

wcwitt:
Can we delete everything python-related (e.g., `- uses: actions/setup-python@v2`) and assume Conda will handle it? Or just the `pip` part?

tjjarvinen:
We need the Python part for PythonCall tests. But we don't need to install any Python packages, as that is taken care by CondaPkg.jl.

wcwitt:
Thanks - let's see what happens https://github.com/ACEsuit/ACE1pack.jl/pull/136.

wcwitt:
The CI succeeded - I'll close.

PR #133, opened 2023-06-23
Fix CI

tjjarvinen:
This should fix CI, by restricting to ACEfit v0.1.0 #132 @wcwitt 

The main point here is that ACEfit v0.1.1 uses PythonCall. ACE1pack has PyCall dependency, from ASE.jl. PyCall and PythonCall should not be dependecy at the same time, as that will cause problems.

For future we need to either remove or change ASE.jl to use PythonCall and thus remove all PyCall dependencys form ACE1pack.

Also, using `dev` option in CI is a bad idea and I took it off. You can test dev-versions on you local comp, by deving to same environment several packages. You can always register new versions to registry to allow CI to use them as a dep.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -20,6 +20,7 @@ jobs:
       matrix:
         julia-version:
           - '1.8'
+          - '1'
           - 'nightly'
         python-version: 
           - '3.8'
@@ -60,9 +61,6 @@ jobs:
           using Pkg
           Pkg.precompile()
         shell: bash -c "julia --color=yes {0}"
-      - name: "Add dev ACEfit"
-        run: |
-          julia --project=@. -e 'using Pkg; Pkg.develop("ACEfit")'
       - uses: julia-actions/julia-runtest@v1
 #  test_mpi:
 #    name: Julia MPI ${{ matrix.julia-version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }} x Python ${{ matrix.python-version }}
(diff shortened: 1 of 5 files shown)


tjjarvinen:
Julia v1.9 and nightly tests seem to fail for sklearn tests. For some reason it pulls ACEfit v0.1.1. I try to disable it with adding Project.toml. But lets see...

wcwitt:
I think this fails to fix the CI because `ACEfit = "0.1.0"` allows anything up to `2.0`. We'd need something like `ACEfit = "< 0.1.0"`.

That would be fine for now, but I would prefer to just fix the problem quickly so that we can continue tagging new `ACEfit` versions and using them in `ACE1pack`.

(Edit: just saw your latest commit.)

tjjarvinen:
That is true! I added `ACEfit = "=0.1.0"` and that should solve it.

tjjarvinen:
> That would be fine for now, but I would prefer to just fix the problem quickly so that we can continue tagging new `ACEfit` versions and using them in `ACE1pack`.

To tag new versions means we need to remove `PyCall` from `ACE1pack`, as we already have `PythonCall` in `ACEfit`.

wcwitt:
I tried that yesterday but couldn't manage to resolve everything. I'll make a PR with what I have and you can take a look.

wcwitt:
The CI is still failing, I think now for a completely different reason, possibly related to introducing a Project file for the tests.

tjjarvinen:
Yep, `Artifacts.toml`, `ACE1` and `ACE1x` missing.  Should work now.

wcwitt:
The CI succeeded, I'll merge. Thanks!

PR #144, opened 2023-07-22
More potential analysis tools

cortner:
Starting to move some of the potential analysis tools I've been using into ACE1pack. 

- [x] rdf
- [x] dimer curves
- [x] trimer curves
- [x] decohesion curves 
- [x] Also TODO: fix CI failure on nightly. This is related to method overwriting and happens to be methods related to *this* PR: 

```
ERROR: LoadError: Method overwriting is not permitted during Module precompile.
```
Easy fix, that would fit well into #144


cortner:
@CheukHinHoJerry -- would you be willing to add the trimer curve and plts to this PR? Can be a rough draft for now. I suspect we will iterate quite a bit on how to best make these analysis tools available to users.

CheukHinHoJerry:
Yes sure - will do this later tonight.

CheukHinHoJerry:
I have no permission to push to this repo directly. Could you please take a look?

Update: Thank you. Just pushed.

CheukHinHoJerry:
An equivalent function of :
```julia
function dimers(potential, elements; 
                rr = range(1e-3, _cutoff(potential), length=200), 
                minE = -1e10, maxE = 1e10, )
   zz = AtomicNumber.(elements)
   dimers = Dict() 
   for i = 1:length(zz), j = 1:i
      z1 = zz[i]
      z0 = zz[j]
      v01 = dimer_energy.(Ref(potential), rr, z1, z0)
      v01 = max.(min.(v01, maxE), minE)
      dimers[(z1, z0)] = (rr, v01)
   end
   return dimers
end
```
would be returning a Dict() swapping the range of given $\theta$, $r_1$ and $r_2$ and different combination of $z$s and that would be a huge Dictionary. Is that fine or I should make $r_1$ and $r_2$ as an arg?

cortner:
this is ready I think, though the smoothness prior tutorial needs another edit by James (separate PR I think) and the new functions ought to be documented ...

ext/ACE1pack_Plots_ext.jl
```diff
@@ -0,0 +1,13 @@
+
+module ACE1pack_Plots_ext
```
wcwitt: @cortner, Can you make a final decision about this before we merge?

src/analysis/dataset_analysis.jl
```diff
@@ -0,0 +1,127 @@
+
+using JuLIP: Atoms, energy, cutoff
+using JuLIP.Potentials: AbstractZList
+
+# z_pairs(obj) = z_pairs(zlist(obj))
```
wcwitt: Any reason to keep this?

src/analysis/dataset_analysis.jl
```diff
@@ -0,0 +1,127 @@
...
+
+# function z_pairs(zlist::AbstractZList)
```
wcwitt: Any reason to keep this?

src/analysis/potential_analysis.jl
```diff
@@ -0,0 +1,209 @@
...
+      z2 = zz[k]
+      v01 = trimer_energy.(Ref(potential), r1, r2, θ, z0, z1, z2)
+      v01 = max.(min.(v01, maxE), minE)
+      trimers[(z0, z1, z2)] = (θ, v01)
+   end
+   copy_zz_sym!(trimers)
+   return trimers
+end
+
+# function trimers(potential, elements; r1 = 0.5:3:_cutoff(potential), r2 = 0.5:3:_cutoff(potential),
```
wcwitt: Should we remove this before the merge?

wcwitt (changes requested):

ext/ACE1pack_Plots_ext.jl
cortner: I will remove this.

cortner: done

src/analysis/dataset_analysis.jl
cortner: removed

src/analysis/dataset_analysis.jl
cortner: removed

src/analysis/potential_analysis.jl
cortner: I wasn't sure which variant of the trimer function to use ... but this might be easy enough to recover with a few minutes' work. so yes I'll get rid of it too.

cortner:
@wcwitt  -- can you please take one more look? I hope this is ready now.

wcwitt:
brilliant, thanks

cortner:
Thanks!

Before we merge your next PR can you please take a close look at the current main and what Ive done to achieve similar things?

If it's all clear to you how to proceed then please just ping me and explain. I'll then review.

Issue #143, opened 2023-07-22
Cleanup Dependencies


cortner:
Can we please remove all dependencies that we don't need. E.g. is MPI really ever used?  I will remove Plots into an extension.

wcwitt:
Is there a Julia tool for identifying dependencies that aren't used?

cortner:
good question. I didn't explore this.  .... and didn't find anythng just now ...

PR #145, opened 2023-07-22
clean up dependencies.

wcwitt:
Addresses #143. I've gone through and removed everything not explicitly used in `src/`. That doesn't mean everything with `using` is _actually_ used, but I didn't take the time to dig that deeply.

The caveat is that many of these dependencies are required for `scripts/ACE1packMPI.jl`, but that is now a bit out of date. We can re-add them as appropriate.

Okay @cortner?


cortner:
How about creating an MPI extension?

Similar to what I'm doing with the plotting extension.

wcwitt:
Yeah maybe - I will consider that when I find time to revive the MPI properly. I decided to remove it completely for now.

Do you know why the CI is failing on Julia nightly? Maybe they've gotten more strict about some of these warnings?

cortner:
I don't. I can try to look into it. Cannot promise when.

cortner:
```
ERROR: LoadError: Method overwriting is not permitted during Module precompile.
```
Easy fix, that would fit well into #144 

I am guessing that to optimize precompilation they are now *forcing* packages to not overwrite anything. Would also help to curb type piracy. But I'm just speculating here.

Issue #147, opened 2023-08-01
Software Paper Revision


cortner:
I'm collecting here some thoughts for potential changes in resonse to the referee reports and/or discussions amongst ourselves: 

- [x] Double check that no documentation pages give 404 errors before final resubmission.
- [x] Finalize Si section : briefly explain choice of smoothness priors, demonstrate smoothness of potentials ... 
- [x] One referee points out that referring to Julia artifacts is confusing. Maybe we should rename `ACE1pack.artifact` -> `ACE1pack.example_dataset` or something similar. Could even be an alias instead of a rename.  (+ revive this in the first place as it got accidentally removed). Suggestion: combine the artifact + `read_xyz` into `read_example_dataset` -> cf #149 
- [x] Molly tutorial ( @tjjarvinen )
- [x] Should the inner cutoff be removed from the software paper entirely? We now never use it I believe? Do any of the examples use it?  -> leave as is ...
- [x] Is it ok to show only the simple Agnesi(2,2) transform as a prototype but use more general variants for all of the examples? -> rephrase as "illustrative" in the article
- [x] Rename `ACE1pack` -> `ACEpotentials`
- [x] Finalize H2O section 
- [x] mention alternative smoothness priors in methods section. (JAMES)
- [x] Tutorials say that lammps export is broken. Is this now fixed? Can those lines be removed? 
- [x] TiAl-basis tutorial is broken due to missing `row_info` in `ACEfit.jl`
- [x] ACE1pack interface tutorial is also broken and out of date. What will we do with that? 

## From previous list of TODOs 

- ~rewrite "intro to ACE models" once the manuscript is done and base it on that.~
- ~link to software paper as intro to ACE~
- ~need a page in the docs on how to prepare data to work with ACE1pack #95~

## ASAP AFTER SUBMIT REVISION

- Make all examples available through ACEworkflows. (unless we want to keep them in the ACE1pack repo)
- the TiAl rdf example in the dataanalysis tutorial is incredibly boring, do we have a nicer dataset to add to the artifacts?  ->  add a small H2O dataset to the artifacts
- Add a function `load_ace_potential` and `save_ace_potential`
- Coordinate these steps with `ACEmd.jl` 

## URGENT FUTURE 

- for future compatibility consider switching ALL models to splines from the start.
- Rename `ACE1Model` -> `ACEpot` or something along those lines... -> maybe this is not urgent, can consider for later.
- Improve documentation of transforms

PR #148, opened 2023-08-07
add Molly tutorial and ACEmd

tjjarvinen:
This will implement Molly tutorial for #147


cortner:
Thank you - your message says "will" but this looks essentially ready? 

My only question is whether `ACEmd` should be a dependency for `ACE1pack` given it is only exported but then not used. How heavy is that dependency? Does it pull in also Molly? Other heavy packages?

tjjarvinen:
yes this is essentially done.

For the export, I was a considering also not exporting. But considering JuLIP retirement, ACEmd is coming in at some point. ACEmd does not pull more than CellListMap, Unitful and AtomsBase in addition to what we have, Molly is an extension.

The other option would be to have ACEmd as a weakdep, so that we can control version in Project.toml, as the tutorial does make some expectations on ACEmd version.

I am not completely sure for the export and now that I think about it, the weakdep route might be a better at this point.

cortner:
Ok thank you, that makes sense. - leave for now and wait for feedback from @wcwitt - I think his preference might be to minimise extensions.

cortner:
You say "Essentially done", what is left?

tjjarvinen:
> You say "Essentially done", what is left?

Only requested changes

cortner:
@wcwitt - unless you have an objection I will merge?

wcwitt:
No objections, and thank you. Longer term, it might be useful to clarify the relationship between `ACEpotential` in ACEmd and `ACE1Model`. For example, I like that ACEmd has a `load_ace_model` routine ... ACE1pack really needs something like that to encapsulate the `read_dict` stuff. But an ACE1pack user who utilizes that routine currently may not get what they expect

cortner:
Thanks, Chuck, I added those points to the TODO list. I'll merge this for now. There will be work to fix ACEmd to correctly interact with some of those changes, but we can do this in future PRs.

cortner:
thank you, @tjjarvinen

PR #149, opened 2023-08-17
WIP: convenient access to example datasets

cortner:
first draft. Things to decide: 
- should we provide metadata, such as references  (currently an option)
- format how it is returned, currently a named tuple, which can also just be splatted into three outputs `train, test, meta`. 
- When only one dataset is provided, then `test = nothing`, is this ok?


cortner:
seems CI is broken, but otherwise this is ready - @wcwitt  if you are still available are you able to take a quick peak whether you are ok with the overall idea?

wcwitt:
Thanks, I like this a lot. Merging and we can fix any issues as they arise.

PR #151, opened 2023-08-18
Possible name change

wcwitt:
I started a fork to investigate the possiblity of a name change. If we move forward, we can either merge those changes here or detach the fork.

Some questions to resolve

* [ ] What to do about the package uuid? https://pkgdocs.julialang.org/v1/toml-files/#The-uuid-field
* [ ] How to update the artifacts?


cortner:
Here is a really nice idea how to manage renaming of packages:  https://discourse.julialang.org/t/changing-the-name-of-a-package/14960/4

Basically the idea is to first fork the old repo, then rename the old repo to the new name but keep the old name in the fork. In the fork we then register a final version of the package that prints a warning that it is now superceded by ACEpotentials.

If you are happy with that, then I suggest
- create a second fork, detach it. Point ACEregistry to that new repo. Check that some of our old scripts still work.
- merge our new fork into the original ACE1pack repo. Rename everything. Register it as a new package in ACEregistry with a new UUID.
- update the ACEworkflows scripts to work with the new ACEpotentials package.

wcwitt:
That looks perfect. To start, I will

* merge these changes into a BRANCH at the original repo (the one we'll keep)
* delete the first fork I created (which currently has the new name)

PR #154, opened 2023-08-19
Renaming continues

wcwitt:
Continuing from #151.

> Here is a really nice idea how to manage renaming of packages: https://discourse.julialang.org/t/changing-the-name-of-a-package/14960/4
> Basically the idea is to first fork the old repo, then rename the old repo to the new name but keep the old name in the fork. In the fork we then register a final version of the package that prints a warning that it is now superceded by ACEpotentials.
> If you are happy with that, then I suggest
> * create a second fork, detach it. Point ACEregistry to that new repo. Check that some of our old scripts still work.
> * merge our new fork into the original ACE1pack repo. Rename everything. Register it as a new package in ACEregistry with a new UUID.
> * update the ACEworkflows scripts to work with the new ACEpotentials package.

Not sure precisely _when_ you want this to happen and I am reluctant to do it now without double checking. So:

* This branch/PR [in the repo we will keep] contains the preliminary name changes to the source code. I will now continue with the parts I haven't touched yet (e.g., the documentation).
* Whenever you are ready, we can merge this, doing whatever is necessary to finalize the name change
* After that we can make a new fork, give it the old name, and revert to a point before the name change. [If I make a new fork now, it would need a temporary name since we can't give it ACE1pack yet.]
* And proceed with the rest of your steps ...


wcwitt:
I've now changed the name everywhere I could find in this repo except for the artifacts; we can handle them separately. Let's see what the CI does.

wcwitt:
To my amazement, the tests and documentation both pass. However, the documentation log is riddled with literate warnings/failures and the docs don't seem to have deployed correctly.

I've checked on `main`, and many of the doc issues occur there too. I will fix those in main, merge into this branch, and see what errors remain.

[Separately, how is the CI passing with all of these documentation warnings (which are actually errors)? We need to change some setting so we are alerted.]

cortner:
Thanks for moving this along.

re docs : I was wondering about running all literate tutorials from runtests in a try block and alert with a failed test if it goes to catch

wcwitt:
That was going to be my last resort (and may still be worthwhile). But I just found a `strict = true` option for the doc build that seems to do what I want.

wcwitt:
I've now merged the tutorial fixes here and the CI passes without weirdness. We can proceed by renaming the repo itself whenever we're ready, I think.

Alternatively, we can continue making changes in `main`, so long as we merge them here.

PR #156, opened 2023-08-19
Fix broken tutorials

wcwitt:
The tutorials all run now (as will hopefully be confirmed in the documentation part of the CI). I've also fixed some other small errors that were preventing the documentation from building fully.

Some of the fixes were ugly, and much of the documentation still requires an overhaul. My goal here is just to get everything running.


wcwitt:
Confirming the tutorials run now (also in CI). I don't think there is any harm in merging, and can catch lingering issues with the documentation in another pass.

PR #157, opened 2023-08-20
Tutorials Testing

cortner:
Here is what I had in mind. It has a potential advantage, namely quick local testing as opposed to needing CI. But I'm not convinced it's needed, we can discuss.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -20,4 +20,6 @@ using ACEpotentials, Test, LazyArtifacts
 
 #    @testset "Test ace_fit.jl script" begin include("test_ace_fit.jl") end
 
+    @testset "Tutorials" begin include("test_tutorials.jl") end
+
 end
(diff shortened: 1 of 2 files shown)


wcwitt:
This will sound unrelated but bear with me: Should we move the tutorials directory into `docs`? Currently, there is a small problem where some tutorials use packages (e.g. Plots.jl) that are not in the main Project.toml. But those packages _are_ in the docs project.

The reason I mention it now is that

> namely quick local testing as opposed to needing CI

can also be achieved by building the docs locally - that's what I sometimes do.

Maybe we should add the full docs build as a test in the manner you describe? Preferably in a way that doesn't also waste CI minutes.

cortner:
I was going to add the relevant packages to the test Project.toml. 

But you are right - we can just build the docs locally too. It’s just an extra step and a small extra barrier. 

But I’m not against your suggestion. Let me think on it?

cortner:
I think we decided to not do this for now, so I'm closing this.

PR #158, opened 2023-08-21
Fix the docs build after renaming

wcwitt:
I think the docs build is failing because of the UUID change, but I'm not sure. Testing now.


wcwitt:
That didn't fix things so I reverted back to the new UUID, deleted ACEpotentials from the docs project, and then re-added it such that the UUID updated.

wcwitt:
That seemed to work, merging.

PR #160, opened 2023-08-22
Wcw/docs

wcwitt:
A few docs-related issues


cortner:
I'll try to consolidate our two PRs ... :..(

wcwitt:
No worries - should be easy to merge. I just have one more thing to do tonight anyway.

cortner:
I rebased and will quickly merge my branch into yours so we don't diverge too much.

wcwitt:
@cortner, you asked for a committee example using the model interface. See this commit: 901af54f2f754679fdd484be735b4ac177d09b69.

Note that I had to update `src/model.jl`. There is very likely a better way, but it's all I could manage for tonight. If you hate it, feel free to revert and we can try something else.

cortner:
**THANK YOU** I'll see what I can do from here.

cortner:
Chuck - I looked at your tutorial draft and I'm actually completely fine with just using co_energy(model.potential). It's the same for just plain old energy, forces etc. We can discuss introducing those for model as well, but I don't think it is crucial for this version of the package.

cortner:
I think this is ready. Please merge if you agree.

cortner:
P.S.: I noticed the `Least Squares Solvers` tutorial is written for the old JSON interface. I thought it would be nice to convert that to pure Julia hence I didn't move it to outdated. But I now don't have time to fix it. Maybe we can add this to our todo list for "asap"?

wcwitt:
> Chuck - I looked at your tutorial draft and I'm actually completely fine with just using co_energy(model.potential).

It's more I thought you might object to what I had to do to `src/model.jl` in order to make that work. But I suppose it should be harmless and and we can always improve the backend later

Issue #168, opened 2023-08-29
Update Molly Tutorial


cortner:
@tjjarvinen  -- can you please update the molly tutorial to use the simplifications you have implemented? Thanks.

wcwitt:
thanks for doing this ... closing

PR #169, opened 2023-08-29
update Molly tutorial

tjjarvinen:
Update Molly tutorial to reflect the new interface

Ref. #168


tjjarvinen:
The error seems to be in `src/literate_tutorials/TiAl_basis.md`, so not in this PR. But the error blocks doc build, so it needs to be fixed before merging this.

cortner:
thank - I'll look into that.

cortner:
the doc build got fixed, but this one still seems to fail? Do you need to update your fork?

tjjarvinen:
The merge is going to main so only main needs to be updated. The issue seems to to be not finding screen for plot generation during doc build, and I have no idea why it only pops up with this PR.

I cant get the doc build to work on my computer at all, so there is some unstable stuff there. No sure would adding this to CI script help

```yml
steps:
      - name: xorg install
        run: sudo apt-get update && sudo apt-get install -y xorg-dev mesa-utils xvfb libgl1 freeglut3-dev libxrandr-dev libxinerama-dev libxcursor-dev libxi-dev libxext-dev
```

and starting Julia with

```yml
run: DISPLAY=:0 xvfb-run -s '-screen 0 1024x768x24' julia --project=docs/ docs/make.jl
```

This should be done in separate PR ofc.

Other than that I can try to get the doc build to work on separate fork, but as far as I see the issue is not in this PR, so not sure how much that would help.

wcwitt:
Would you please try merging `ACEsuit/main` into this branch? I think that will allow the CI to succeed.

One subtle point when building the docs locally: if there are unregistered changes in ACEpotentials (like now), one needs to `dev ..` from the docs project to get those changes.

This PR worked for me locally after I did both things above.

wcwitt:
Thanks a lot - this looks good now. Okay to merge?

PR #171, opened 2023-08-30
Fix docs build

cortner:


wcwitt:
Thanks, I was just wondering about this

cortner:
Some exports changed somewhere. I'm on it. Will try to fix today.

wcwitt:
Oh it looked to me like your commit fixed it already.

Also I was hoping we’d be able to test the versioned docs here based on the
tag you made. But it looks like we’ll have to merge before we’ll see the
versions

cortner:
indeed it did, I thought I saw a failure, but it must have been a different branch, then I can merge now.

Issue #35, opened 2022-07-18
Simplify and document generation of (per-site) descriptors


wcwitt:
Quick summary of a Slack discussion from 16-18 July 2022.

"before I fit a potential I would like to just evaluate the basis for an atomic structure, i.e. get an array of basis values for each site. Can someone please put such an example into this tutorial?"

Options mentioned
* `site_energy(B, at, 1)` where 1 is atom index, at is atoms object and B is ACE basis
* `evaluate_basis(B, at)`

There was some disagreement about whether `site_energy(B, at, 1)` is adequate or confusing. It provides the energy basis evaluated for a site, and the total site energy for a potential `IP` is something like
```
site_energy(IP, at, 1) = sum(coeff .* site_energy(B, at, 1))
```

One thought: Would `site_descriptor(B, at, 1)` be more clear than `site_energy(B, at, 1)`? More generally, we need to discuss, decide whether to make any changes, and document.

cortner:
`evaluate_basis(basis, args...)` doesn't work for me. The fact it's a basis we are evaluating is already encoded in the argument. From the discussion on Slack I suggest: 
* `evaluate(basis, atomic_env)` : the current evaluate
* `evaluate(basis, structure)` : the current `energy`
And then, `energy` would just become an alias. 

Next - do we need an alias for `evaluate` that sounds more like "evaluate the descriptor"? The `site_descriptor` could just be achieved via 
```julia
evaluate(basis, at, 1)
```
Instead of additional syntax for evaluating *all* the per-site descriptors we could allow 
```julia
evaluate(basis, at, [2,3,5,10]) # for a few site descriptors
evaluate(basis, at, :)  # for all site descriptors.
```
But maybe this is getting too cryptic now and the `site_descriptor` and `site_descriptors` syntax will be clearer. (Or could again be aliases...)

wcwitt:
I added a very basic tutorial addressing this a while ago: https://acesuit.github.io/ACE1pack.jl/dev/literate_tutorials/descriptor/.

So I vote we close for now and revisit when necessary?

cortner:
what about the terminology, that's not really settled yet.

cortner:
@wcwitt  --  A few people at UBC that are starting to use ACE as features for classification and data analysis tasks. So I'd like to come back to this, make this functionatlity convenient to access and then close this. 

[...]

PR #173, opened 2023-09-04
Descriptor

wcwitt:
First draft towards addressing #35.

@cortner do you prefer `site_descriptor` or just `descriptor`? Is there anything you'd like to see changed?


wcwitt:
Maybe this should go in `analysis/structure_analysis.jl` instead?

cortner:
I think it's `site_descriptor` or `site_descriptors` (plural). `descriptor` is insufficiently precise I think.

wcwitt:
Sounds good, I’ll leave things as they are then.

src/descriptor.jl
```diff
@@ -0,0 +1,21 @@
...
+
+Compute the site descriptor for the `i`th atom in `atoms`.
+"""
+function site_descriptor(basis, atoms::AbstractAtoms, i::Integer)
+   return site_energy(basis, atoms, i)
+end
+
+
+"""
+    site_descriptor(basis, atoms::AbstractAtoms)
```
cortner: docs need plural

src/descriptor.jl
```diff
@@ -0,0 +1,21 @@
...
+
+Compute site descriptors for all atoms in `atoms`, returning them as a list.
+"""
+function site_descriptors(basis, atoms::AbstractAtoms)
+    return [site_descriptor(basis, atoms, i) for i in 1:length(atoms)]
```
cortner, line 21: should it be a vector of vectors or a matrix? I guess I'm ok with either for now and we could just see what people who actually use this will say?

cortner (commented):
sorry don't have time to make changes right now. Here two quick comments - can also be ignored - please merge anytime.

src/descriptor.jl
wcwitt: done, thanks

src/descriptor.jl
wcwitt: Good point - I'm not sure what's best. Will leave for now

PR #182, opened 2023-10-11
AtomsBase fitting support

tjjarvinen:
This adds support to use AtomsBase data with acemodel interface.

JuLIP interface will work as it used to be. I did not change anything on it. You can force new assemble function, that is in ACEmd as an extension to ACEfit, with JuLIP too by giving `new_assembly=true` keyword.

I did not implement repulsion_restraint functionality for AtomsBase - for now.

Note! This is not final yet, as it lacks tests. But it is a basis for discussion on what needs to be added still.


docs/src/tutorials/AtomsBase_interface.md
```diff
@@ -0,0 +1,130 @@
...
+```julia
+# return weight for a structure 
+data_point[:weight]
+```
+
+To set a weight by hand on an individual structure you can use
+
+```julia
+# Set weight
+data_point[:weight] = 60
```
cortner, line 88: please add an example how to change individual force or energy weights.

docs/src/tutorials/AtomsBase_interface.md
```diff
@@ -0,0 +1,130 @@
...
+ACEpotentials.linear_errors(data_train, model; weights=weights);
+```
+
+### Weights with AtomsBase
+
+With AtomsBase weights are stored in AtomsBase structures and there is no need to create `AtomsData` structures. Weight for a structure can be accessed with keyword `:weight`
+
+```julia
+# return weight for a structure 
+data_point[:weight]
```
cortner, line 81: please explain how this per-structure weight modifies the E, F, V weights.

cortner (changes requested):

tjjarvinen:
This PR would be ready for merge.

But docs seem to fail because of `strict = true` option  in `docs/make.jl` causing

```julia
ERROR: LoadError: ArgumentError: makedocs() got passed invalid keyword arguments:
  strict = true
```

How do we deal with this, just remove it, or something else?

cortner:
@wcwitt  -- I'm quite keen on this PR since it will allow us to slowly move away from JuLIP and integrate better with the rest of the Julia ecosystem. It shouldn't change any default behaviour at all. Would you mind taking a quick look to make sure you are happy? 

Do you have a view on the docs issue?

wcwitt:
Thanks! I'm finished teaching after today - will look closely ASAP.

tjjarvinen:
@wcwitt do you have eta for comments on, so that we can move forward with it?

wcwitt:
Sorry, probably tomorrow.

tjjarvinen:
I removed `strict=true` option form docs, as it had been removed from Documenter (ref. [here](https://github.com/JuliaDocs/Documenter.jl/blob/v1.0.0/CHANGELOG.md)). The behaviour is not changed, as by default docs fail, if doctests don't pass.

wcwitt:
Sorry to be slow on this. One reason I've been cautious is that I would prefer we consolidate on a single assembly routine before merging.

My main concern with the new assembly has been that the majority of the speedup comes from the fact that it doesn't garbage collect after each config. That's fine for toy examples but - at least historically - the manual garbage collection has been crucial for avoiding out-of-memory problems in real world fits.

To that end, I've been testing the various options currently living in `ACEfit/dev-assemble`. Strangely, I haven't been able to reproduce the out-of-memory issues at all. Which makes me wonder if something was fixed for Julia 1.9.

tjjarvinen:
My main idea with the assembly was that we could test the new version here. While  having a fallback for the old assembly in hand, if there would be some issues. But Christoph wanted to have the new assembly available for other ACE use cases too, hence the PR in ACEfit.

The origin for the memory leaks/issues is a concerning matter. Do we know what package causes it? Is it ACE1, maybe data transfer related (SharedArrays) or something else? I did stumble to this [BUG](https://github.com/ACEsuit/ACEmd.jl/issues/47) when implementing assembly, but that might be completely another issue.

cortner:
https://discourse.julialang.org/t/julia-1-10-rc1-gives-a-huge-speed-up-over-julia-1-9/106072

wcwitt:
Since this doesn't break any current functionality, we can merge for the sake of moving forward. It seems clear we'll need a big breaking change soon that drops JuLIP entirely, and we can argue about the API/etc then if necessary

cortner:
Great. @tjjarvinen - anything left to consider from your end?

wcwitt:
The CI just passed, meaning the merge conflicts seem resolved

tjjarvinen:
Everything should be fine now from my end.

docs/src/tutorials/AtomsBase_interface.md
tjjarvinen: done in documentation

cortner:
Can you please add a sentence to the top of the docs that this interface should currently be considered experimental.

tjjarvinen:
[There](https://github.com/ACEsuit/ACEpotentials.jl/blob/main/docs/src/tutorials/AtomsBase_interface.md) a comment of being not rigorously tested yet. Saying that:

"AtomsBase interface has not been rigorously tested yet. So, expect some issues here and there. But you are recommended to give it a try."

tjjarvinen:
I will look for the docs, once CI and doc deploy finishes and fix some typos and look, if something needs to be added.

Issue #167, opened 2023-08-25
Save and Load Potentials


cortner:
Add a function `load_ace_potential` and `save_ace_potential`.  This needs to support typical usage from python, Julia, and in particular ACEmd. 

The main question is whether to store only the fitted calculator, or the entire `ACE1Model`, which includes the calculator as well as the basis from which it was fitted.

@tjjarvinen  --  do you have a view on that?

tjjarvinen:
Can we evaluate the basis without ACE1 or can we move the basis to some other programs? If not, then I my opinion we should not save the basis.

cortner:
The basis and potential are evaluated by the same codes. ACE1 right now - something else in the future.

cortner:
I think I’m going to make it an option then we will see what is actually needed.

cortner:
I'll add another TODO to this issue that @bernstei asked for: The entire manifest should be part of the potential to guarantee reproducable results.

tjjarvinen:
Ok, that would mean saving manifest together with the potential or at least version information about ACE1 to the potential file.

There is also an other option - using containers. This is possible with the new ASE calculator that I now have working. This should be a lot more better option than saving manifest. And also make using (installing) ACEpotentials a lot more easier for people not familiar with Julia.

cortner:
let's please discuss this in our next meeting. There is the immediate need of producing ACE potentials and using them in Python. But separately we will produce many other models for different problems in the future. And how to manage that?

wcwitt:
@tjjarvinen, would you be willing to adapt a version of the potential loading routine from ACEmd and put it in ACEpotentials? It looks to me that what is needed is essentially a copy/paste of your code, removing the last bit that returns a new-style potential.

tjjarvinen:
Yes I can do that. Do you have preferences for name?

Just to add note, calling `ACEmd.load_ace_model` with keyword argument `old_format=false` would do the same thing already. I am now presuming this is related to the Slack discussion with Noam? That option is there to implement tests between old JuLIP style potentials and new AtomsBase.

So, one option would be, to just make that call a bit more explicit that you can load JuLIP style potentials too.

cortner:
I'm closing this in favour of a new discussion on how to save and load models, I will start a separate Issue #217

PR #183, opened 2023-10-27
Add updated save/load functions

tjjarvinen:
This is  related to #167 

- Add information about packages to the potential save file
- I made saving simpler - no need to use `write_dict` etc. anymore
- loading is also simpler and has option for `JuLIP` or `ACEmd `styles
- Added a new page on docs for saving and loading.

I am not sure about the names `save_ace_potential` and `load_ace_potential`. But `ACE1` exports `save_potential` and `load_potential`, which are defined in away that makes extending very difficult. And thus I used new names.

The version information saved now should make it possible to build a system where the potential should work. But, this of course will be something that only time will tell.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,7 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
 
 [compat]
-ACE1 = "0.11.15"
+ACE1 = "0.12"
 ACE1x = "0.1.8"
 ACEfit = "0.1.4"
 ACEmd = "0.1.5"
--- a/Project.toml
+++ b/Project.toml
@@ -13,11 +13,11 @@ JuLIP = "945c410c-986d-556a-acb1-167a618e0462"
 LazyArtifacts = "4af54fe1-eca0-43a8-85a7-787d91b784e3"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
-
 
 [compat]
 ACE1 = "0.11.15"
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -64,6 +64,7 @@ makedocs(;
         "Home" => "index.md",
         "Getting Started" => Any[
             "gettingstarted/installation.md",
+            "gettingstarted/saving-and-loading.md",
             "Tutorials" => Any[
                 "tutorials/index.md",
                 "literate_tutorials/first_example_model.md",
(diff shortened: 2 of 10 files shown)


tjjarvinen:
For a reference this is example version info that is printed on potential load

```
This potential was saved with following versions:

JuLIP v0.14.5
ACEbase v0.4.3
ACE1x v0.1.8
ACE1 v0.11.16
ACEmd v0.1.7
ACEpotentials v0.6.3
ACEfit v0.1.4

If you have problems with using this potential, pin your installation to above versions.
```

tjjarvinen:
Doc build fails because of `strict=true` keyword on `makedocs`. This is not supported by current versions of Documenter, as described here https://github.com/JuliaDocs/Documenter.jl/blob/v1.0.0/CHANGELOG.md So I removed it.

wcwitt:
It seems when they removed the `strict` keyword, they also changed the default to be consistent with the old `strict=true`. So I'm happy with that.

wcwitt:
Thanks a lot for this @tjjarvinen.

> I am not sure about the names save_ace_potential and load_ace_potential. But ACE1 exports save_potential and load_potential, which are defined in away that makes extending very difficult. And thus I used new names.

I'm not certain either, but I think it's worth discussing a bit. What about simply `load`?

Also, we already have `export2json` and `export2lammps`. In at least one place, you replaced `export2json` with `save_ace_potential`? Do we really want that? [Edit: to be clear, I don't have a strong view. but...] I would prefer to have one way of doing things.

CC: @cortner

cortner:
I thought it was a just a way to add some functionality without changing the behaviour of the current code and not to replace it / duplicate it. Going forward we need to revisit this anyhow and have a discussion how we will save and load potentials.

cortner:
I see now that this actually makes an attempt to already make the updates I allude to above. I'll need more time to review. 

At first glance it looks like it is doing the right thing but I agree we need to fix the naming and there is no reason to not keep our `export2json` and maybe make up an import `importjson`

wcwitt:
I guess my point is that we _absolutely_ need a better way to load potentials. (I was just in Oxford and someone complained it took them an hour to figure that part out.) But not sure about multiple equivalent options for saving.

Edit: I wrote this without seeing your last post. I agree with it.

cortner:
Btw, the old `ACE1` functions can  just be deleted. We can make a breaking version bump in `ACE1` that just gets rid of them and bring them into `ACEpotentials`.

cortner:
Ok ... just on the naming ... I'd be happy with either 
- `save_potential`, `load_potential` 
- `export2json`, `importjson`

it actually feels like the first option is cleaner. So why not just kill the versions in ACE1? I'm unsure about the long name `save_ace_potential` but I don't feel too strongly about that.

wcwitt:
> it actually feels like the first option is cleaner. So why not just kill the versions in ACE1?

Works for me.

cortner:
we tagged a new ACE1 v0.12.0, this PR should rename to `load_potential` etc and require ACE1@0.12

cortner:
Thank you, Teemu. What about deprecating `export2json`?

tjjarvinen:
That is a good point. Is using `@deprecate` ok?

cortner:
I think so. I cant think a reason not to?

cortner:
so now there is only one more question for me -- should there be a `metadata` kwarg for `save_potential` to maintain compatibility with `export2json`? I have no idea whether anybody used this?

tjjarvinen:
I added `meta` kword. But could change it to `metadata` too, if that is better.

cortner:
I’m ok either way

cortner:
@wcwitt  -- from my end this is ready.

wcwitt:
Sorry - I missed this ping. Thanks @tjjarvinen!

Issue #188, opened 2023-11-14
New ACE Kernels


cortner:
As a first step towards retiring ACE1 entirely, I'm putting together a new evaluator (currently in a private repo). For now, it does the following: 
- evaluate radial basis via SVector splines 
- sphericart spherical harmonics 
- pooling via P4ML
- StaticPolynomials for the n-correlations
-  import ACE1 potentials into the new format (correctly!)

This next issue is to incorporate this here, provide wrappers for both JuLIP and AtomsBase (Molly), and move then discuss how to move over the entire ecosystem, I'm collecting some todos:  
- [x] switch ACE1 over to using splines for the radial basis so that fewer points can be used
- [x] also implement the entire basis sets with the new kernels
- [ ] bring the purification to the new setting

cortner:
The evaluator is ready to incorporate here. I'd like to get feedback on the following question: Should it continue to live in a separate repo, or should I move the code into `ACEpotentials.jl` with the idea that it will localize the development, maybe simplify maintenance going forward, and be less confusing for outside users? 

CC @casv2 @wcwitt @tjjarvinen

wcwitt:
I would be happy if we started to localize, but I also don't feel super strongly.

tjjarvinen:
So it has the evaluator and now needs pairlist and atomic structure data input?

To aswer I would need to know a more details on what kind of input it takes etc.

For the start I would not pull it in ACEpotentials. I would first at least do some preliminary implementation for JuLIP and AtomsBase first before PR. Also, how does this stand with JuLIP retire plan?

cortner:
The new evaluator doesn't need JuLIP but can get a JuLIP wrapper.

cortner:
Structure and pairlisz can be anything. Needs a wrapper though.

tjjarvinen:
Then ACEmd might be a better place for AtomsBase wrapper.

Does it calculate per atom inputs or does it take a group atoms at once?

cortner:
at the moment it does per-atom. I want to start experimenting with groups of 16 or 32, ...

tjjarvinen:
per-atom implementation is very easy just change the ACEmd backend functions https://github.com/ACEsuit/ACEmd.jl/blob/main/src/backend.jl and after that you will have fully working AtomsBase implementation.

For groups we need to change more things, but it should not be much.

cortner:
The wrapper should be easy enough. it's the kernels that concern me.

[...]

PR #189, opened 2023-11-24
UF_ACE - JuLIP Interface

cortner:
This is a first step towards #188 . So far all it can do is take a fitted ACE1 / ACE1x type model, then convert it to the new format. The usage is seen in `test/test_uf_ace.jl`. Next step will be to implement the basis as well. I want to discuss this with @CheukHinHoJerry  and @wcwitt  before charging ahead.


cortner:
This is ready to merge. It doesn't change any behaviour, just adds new functionality via 
```julia
ACEpotentials.Experimental.fast_evaluator
```
so I don't think there is any harm in merging. Over the course of the summer I want to integrate this code fully into ACEpotentials and remove and retire UF_ACE.

cortner:
I now added a tutorial to demo the fast evaluator.

PR #194, opened 2024-05-02
bugfix.

### LINKED ISSUE #193: Tests Failing
Test tests are currently failing - in a seemingly begnin way. Need to go through them and adjust/fix as appropriate.

wcwitt:
Aims to resolve #193. Hopefully the CI will pass.

### DIFF
--- a/test/atomsbase.jl
+++ b/test/atomsbase.jl
@@ -24,5 +24,5 @@ using Test
 
     acefit!(model, data_train; solver=solver, weights=weights, prior = P, repulsion_restraint=true);
     ce, err = ACEpotentials.linear_errors(data, model; weights=weights);
-    @test err["mae"]["F"] < 0.5
+    @test err["mae"]["F"] < 0.6
 end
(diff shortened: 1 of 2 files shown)


wcwitt:
The first problem I noticed is fixed, but there is another issue related to `atomsbase`.

wcwitt:
@tjjarvinen would you mind taking a look here?

tjjarvinen:
It seems to be a tolerance issue. The new fitting method, does not work as well as the old one and that tripped one of the (very loose) test tolerance checks. It should work now.

cortner:
Im a bit nervous about loosening the tolerance. What changed?

tjjarvinen:
The change to fitting method does not seem to converge as well as the previous setting.

We could look more on why it does not converge as well or loosen the tolerance. Latter is faster, but we could miss also something.

Also it looks like that ultrafast ACE has some test failing also...

cortner:
re uf_ace, probably just a matter of switching from max-error to mae or even better fixing a seed. Anyhow this is experimental.

cortner:
what do we do about this now?

cortner:
so it was just a random fail. I'll fix that at some point ...

wcwitt:
Sorry for the slow reply - if it fails again we can look more closely.

PR #195, opened 2024-05-03
Prototype ACE models based on P4ML

cortner:
This PR provides experimental implementations of ACE models to provide a fast but flexible implementation of a variety of ACE models to be tested, including trACE-like models, different tensor formats, out nonlinearities (e.g. FS). At the same time it provides a blueprint for implementing non-trivial models using the kernels provided by P4ML.

### Main Items

- [x] transfer most important envelopes from ACE1.jl
- [x] transfer most important transforms from ACE1.jl
- [x] implement learnable radial basis
- [x] implement splined radial basis (non-learnable)
- [x] interfaces for sphericart, and P4ML kernels
- [x] a first ACE model with learnable radials / trace capabilities
- [x] gradient w.r.t. positions
- [x] gradients w.r.t. params
- [x] grad_params_forces (reverse-over-reverse)
- [x] basis(energy)
- [x] basis(grad)
- [x] jacobians 
- [x] test correctness of energy, forces, virials
- [x] test correctness of pullback_EFV
- [x] Pair potential
- [x] an example of linear parameter estimation (regardless of accuracy) 
- [x] an example of nonlinear parameter estimation (regardless of accuracy) 
- [x] evaluating some energies, forces, virials with the fitted potentials (even if crap)
- [x]  The standard Lux printout doesn't show the correct number of parameters. This is probably because I store the weights as arrays of arrays rather than high-dimensional tensors. We can revisit this.

### TODO list towards achieving feature parity

- [x] the tensor operations A -> AA -> B should be wrapped into a separate type rather than hard-coded. This will give better structure and allow easier experimentation with different formats. The question is whether to leave A outside or inside.
- [x] adapt the architecture to enable different tensor formats for A and AA
- [x] Give the Rnlzrr basis the option of freezing the coefficients, and hence making the model linear. (=> achieved via splinification; maybe there are better options.)
- [x] E0s
- [x] Fix the tests
- [ ] The radial basis currently has maxn = maxq, which is of course ridiculous, but it doesn't prevent functionality testing so I'll leave it for now. 
- [ ] The pair potential basis currently uses random trace-like mixing. Should implement a one-hot initialization of the rbasis weights and this could be taken as the default for the pair basis.
- [ ] Give the Rnlzrr basis the option of initializing in different ways, in particular just using the polynomials (which only makes sense for a single species...)
- [ ] purify 2b
- [ ] Outer nonlinearities such as FS, MLP, or a BOP-like thing.
- [ ] general rational transforms, both frozen and learnable, to unify everything we have. Explore `Polynomials` and `StaticPolynomials` packages to manage the algebra.
- [ ] Implement `sparse_ace_spec` as commented in `models/utils.jl`
- [ ] How to incorporate smoothness priors. Ideally we should just incorporate them through a rescaling the A basis (or possible the Rnl, Ylm bases)
- [ ] pushforwards of main kernels for faster jacobians and basis_derivatives
- [ ] weird stuff going on with units. to be discussed and cleaned up. Should we enforce usage of units? Could prevent mabny bugs!
- [ ] writing models to and from files...
- [ ] rewrite E0s to allow general Vref, probably best by utilizing a parameterized stacked potential rather than doing that stacking internally. That could increase flexibility and reduce boilerplate code.
- [ ] discuss whether we ever want to allow complex SH again?


cortner:
@jameskermode  --  FYI, this is becoming a first draft of what I was planning. There are a few steps missing before I would appreciate help with testing and polishing the user-interface. But I'm not very far away. 

In general, this is slowly ready for initial feedback if anybody wants to look at it.

cortner:
@tjjarvinen and @TSGut  -- the new implementation of ACE models is an opportunity to try out more efficient batching. As soon as you have a neighbourlist ready that can provide atomic environments in batches of 16, 32, ..., we can write an alternative evaluator and see how much we gain.

cortner:
@tjjarvinen  -- I think this is also an opportunity to switch to Bumper.  Maybe take a look at this PR before our next meeting.

cortner:
@CheukHinHoJerry  -- I implemented the approach that Avik Pal suggested for reverse-over-reverse. It's stunning how well this works. Take a look at `pullback_2_mixed' in `models/ace.jl` when you have a moment. The performance is also fantastic.

wcwitt:
@cortner when appropriate I'd like to contribute some further tests or examples to get a feel for this new stuff. Would it make sense to start by converting [this tutorial](https://github.com/wcwitt/ACEpotentials-Tutorial), which is long overdue to be moved to this repo? Do you have a better idea?

jameskermode:
Updating/testing tutorials is also something I/others in my group can help with

cortner:
@wcwitt -- I think that would be a great start. Let me finish the three remaining bullet points. At that point we should have feature parity with ACE1 - purification + much more incl trACE-like and learnable radials. 

One question I'd like your feedback on (CC @jameskermode) is whether we should move everything to `AtomsBase` + `AtomsCalculators` or whether it is better to stick with JuLIP for a little while.

jameskermode:
I haven't really kept up to date with AtomsCalculators but as far as AtomsBase is concerned I'd be supportive of making the switch now rather than having to re-engineer ACEpotentials again in the near future.

cortner:
> Updating/testing tutorials is also something I/others in my group can help with

I will need a bit more help than that. I will also need help testing different parameter estimation approaches, different tensor decompositions, random vs learnable radials, regularity priors for learned radials, nonlinear embeddings (FS, MLP, BOP, ...) and more. There are exponentially many variants in my head and I'm not sure where to go with that. I'll document that at some point and we can then discuss how to go about this.

thomas-rocke:
@cortner I'm happy to help do some benchmarking for the parameter estimation, once docs are roughly there.

wcwitt:
> is whether we should move everything to AtomsBase

Whatever you and @jameskermode think. I'm giving a talk immediately after Rachel soon (4 June) - will mention if this is the decision

image

jameskermode:
> @cortner I'm happy to help do some benchmarking for the parameter estimation, once docs are roughly there.

thanks @thomas-rocke - yes, Tom and others in my group will be very happy to explore these parameter options.

cortner:
It's definitely happening. My question is only whether we should try to force the issue by not even providing the option of using JuLIP anymore.

jameskermode:
Could we try it and see what we miss?

tjjarvinen:
To add status on AtomsBase and  AtomsCalculators.

- ACEmd is working with AtomsBase and AtomsCalculators, with ACE1 backend
- I have ASE interface for AtomsCalculators working (tested with ACEmd), meaning ASE interface is ready
- AtomsBase training relies on an ACEfit extension that is located in ACEmd - this need to be moved to either to ACEfit or to where ever the new ACE code resides

Pair potential still uses JuLIP in the background or has that been switched to new AtomsCalculators? Same for one body potential?

CheukHinHoJerry:
Maybe there should be a separate PR for totally migrating to AtomsBase? I think we should at least advertise it a bit more before it happens. 

Also, it sounds to me that there are quite a lot of utilities for fitting/IO in/depends on ACEmd right now (e.g. new ASE interface, save_ace_model …). Should a majority of those be moved to ACEpotential in the near future?

CheukHinHoJerry:
Please let me know when it is time to move the purification code. I could help with that. 

@wcwitt Maybe I can also help with writing some examples if you don’t mind? I will have to get familiar with the new kernel too.

cortner:
> Please let me know when it is time to move the purification code. I could help with that.

I think it may be unrealistic to provide the full purification for this setting. But we should implement the purify-2B only. Not quite ready for it, but relatively soon.

cortner:
@jameskermode @wcwitt  --  AtomsBase forces units on everything. As a matter of principle I find this problematic. I want to be allowed to work with non-dimensional models sometimes. And it's a pain when designing a good interface. 

But for `ACEpotentials.jl` this may well be a good thing. I don't know. It has some interesting consequences. E.g. training weights must get units now otherwise you can't add up loss terms.  E.g. this calculation 
```julia
wE^2 * E^2 / length(at)  
  + wV^2 * sum(abs2, V) / length(at) 
  + wF^2 * sum(_norm_sq, F)
```
if `E, V` have units angstrom and `F` has unit `eV/A`, will only work if `wE, wV` have units `1/eV` and `wF` has unit `A/eV`.  It's also interesting that this shows clearly that we are adding up things that shouldn't really be added (this has always bothered me...)

I also wonder whether this could help avoid unit conversion bugs in the future.

tjjarvinen:
ACEmd implementation that also uses AtomsBase does not use units for training weights. Basically you can just strip the units in the first step and then you are good to go unitless.

The big plus with units is of course is that units force correctness checks into the code and this can be a good thing.

cortner:
I hate using `ustrip` all over the place. It feels like a hack and makes code unreadable. I claim that if that is necessary then there is a problem with the overall design.

cortner:
In trying to fix the unit tests I modified the `test` environment and now get the following: 
```
(ACEpotentials) pkg> test
     Testing ACEpotentials
ERROR: can not merge projects
```
Why do we even need the test environment? Why can't we just work with a single Project.toml? I find the separate test environment annoying to work with, since everytime I try to `activate parent...` in VSCode in a test file it activates the test env instead of the root env which I actually want?

cortner:
@jameskermode , @wcwitt  and anybody else who would like to contribute.  This is far from done, but it is now at a stage where anybody can jump in and help.  I cannot spend my summer bringing this to feature parity with original ACEpotentials, but I can certainly help. If there is enough interest to commit to this, then now is the time to discuss how to proceed.

wcwitt:
> Why do we even need the test environment?

I think this is recommended when the tests have additional (heavy) dependencies. I can't recall at this point, but it may have been a @tjjarvinen contribution, so he may have a view

cortner:
I am wondering how to organize the next steps of development. 

(1) ignore all version control and continue the current branch as a development branch without versions.

(2) just merge into main and continue to develop new experimental features without exporting 

(3) Create a new branch `v0.6` off of `main` where we do only maintenance / bugfixes, then merge the PR into `main` and tag it as `v0.8.x` versions (and eventually 0.9.x and so forth).

Any views on this?

cortner:
@tjjarvinen  - thank you for fixing the two examples at the end of the tutorial. Can you please explain why this substitution resolved the problem with the optimizer? 
```julia
#at_new = AtomsBuilder._set_positions(at, reinterpret(SVector{3, Float64}, results.u) * u"Å")
at_new = FastSystem(
   bounding_box(at),
   boundary_conditions(at),
   reinterpret(SVector{3, Float64}, results.u) * u"Å",
   atomic_symbol(at),
   atomic_number(at),
   atomic_mass(at)
)
```

tjjarvinen:
No it did not fix it. I just didn't have your version of AtomsBuilder, so the commented line did not work for me, and I had to have something else. So, I made the same using standard AtomsBase.

tjjarvinen:
> I think this is recommended when the tests have additional (heavy) dependencies. I can't recall at this point, but it may have been a @tjjarvinen contribution, so he may have a view

Yes this it is mainly the reason. It is a recommendation from Julia v1.2 onwards. The old way of not having test/Project.toml is deprecated and will be removed with Julia v2 (or that is what I understood where we are going).

tjjarvinen:
For the development, I would create a tag on the main branch before we merge anything. In fact we have no tags for release versions at all. It would be really good to create them.

I am in favor of branching `v0.6` and continuing the development in main branch. For this we need to make sure we have documentation for `v0.6`. To this we need to tags, so that documenter saves the documentation in different folders. This is very important!

jameskermode:
Great, I will start playing with this! First step for me will be to confirm I can run the tutorial, and then I'll try to expand towards what is needed for increasingly real applications. I'm happy with branching out a `v0.6` branch and continuing this development in `main` as @tjjarvinen suggests. Once things stablise we can then make a new release.

jameskermode:
Where is a good place to coordinate activity on this? Happy to use this PR, or we could setup a Slack or Zulip channel instead (with a preference for Slack for me).

After some faffing around with dependencies (SparseArrays, StaticArrays, AtomsBase missing from `doc/src/newkernels/Project.toml`, I'll check in a fix) I got the example running. After 3000 Adam steps for the non-linear opt the loss and gradient are still rather large:

```
  3000     1.001428e+07     1.763535e+08
 * time: 5545.555037021637
 ```

but this is sufficient to get more reasonable MAEs after the linear fit:

```
[ Info: MAE(train) = 0.06140387518964288
[ Info: MAE(test) = 0.08700835512433414
```

The MD still explodes.

I am wondering if it would make sense to do a linear fit of the expansion coefficients *before* the non-linear optimisation, and start the non-linear opt from there with the aim of restrict it to improving the radials and species embedding. Of course this risks getting stuck in local minima and losing the benefits of SGD/Adam.

jameskermode:
> I'll check in a fix

@cortner I don't have write access to this repo, could you add that? Or I can do a PR into this one if you prefer.

cortner:
now done. sorry for the delay.  I'll wait for your commit and will then merge this into main as discussed so we can manage further improvements through a sequence of PRs.

cortner:
> I am wondering if it would make sense to do a linear fit of the expansion coefficients before the non-linear optimisation, and start the non-linear opt from there with the aim of restrict it to improving the radials and species embedding. Of course this risks getting stuck in local minima and losing the benefits of SGD/Adam.

I think there are quite a few options here worth exploring. Note also that I'm unsure how good a dataset TiAl actually is. I think we should switch to something else. E.g. I tried a linear fit to this with our tried and tested ACE1 and got complete crap. If anybody wants to suggest a sequence of good datasets that we trust? I am thinking e.g. 
- Cuo et al 
- PRX Silicon and maybe Ralf's Cu
- TiO2 (Artrith, 2016), I tried this with linear ACE and got poor results
-  Anybody have a 3-4 component alloy? 
- the "classical" molecular benchmarks rMD17, 3BPA etc?

CheukHinHoJerry:
> > I am wondering if it would make sense to do a linear fit of the expansion coefficients before the non-linear optimisation, and start the non-linear opt from there with the aim of restrict it to improving the radials and species embedding. Of course this risks getting stuck in local minima and losing the benefits of SGD/Adam.
> 
> I think there are quite a few options here worth exploring. Note also that I'm unsure how good a dataset TiAl actually is. I think we should switch to something else. E.g. I tried a linear fit to this with our tried and tested ACE1 and got complete crap. If anybody wants to suggest a sequence of good datasets that we trust? I am thinking e.g.
> 
> * Cuo et al
> * PRX Silicon and maybe Ralf's Cu
> * TiO2 (Artrith, 2016), I tried this with linear ACE and got poor results
> * Anybody have a 3-4 component alloy?
> * the "classical" molecular benchmarks rMD17, 3BPA etc?

@YangshuaiWang and I are playing with 2-5 component alloy and also 3BPA.

jameskermode:
Changes pushed, thanks.

I can also add a 3-component FeCrNi alloy (or 4-components including spin up and spin down Fe as two species) from [this recent paper](http://dx.doi.org/10.1103/PhysRevMaterials.8.033804). We tried and failed to fit a linear ACE to this without trACE compression due to memory limits.

cortner:
This is now merged. Thank you to those who were following. I will file a separate issue outlining the long TODO list which will be addressed in separate PRs, and where we can also discuss. 

Just to be 100% clear, I do NOT expect the current version to fit a useful potential out of the box (see also James' comment about 3000 Adams iterations). The next few weeks / months will be to fine-tune the codes and defaults to make that possible.

PR #197, opened 2024-05-22
WIP: Experiments with Bumper

cortner:
This PR is to adapt the new model kernels to use Bumper in order to avoid allocating temporaries. As a general note, we also have `ObjectPools` for this. This PR is an experiment to replave ObjectPools with Bumper so we can stop maintaining it.


cortner:
Not a huge gain in performance (though I also fixed some smaller type stabilities along the way) but the allocations are down quite a bit. I believe the remainder are sphericart allocations which I can fix next. 
```
[ Info: old evaluate
  24.583 μs (75 allocations: 115.59 KiB)
[ Info: with bumper
  18.833 μs (3 allocations: 1.81 KiB)
```

cortner:
this is now obsolete and will be replaced by the latest development branch of P4ML and WithAlloc. I'll close this now.

PR #198, opened 2024-05-27
Update documentation for AtomsBase

tjjarvinen:
AtomsBase documentation is a bit out of date. This will update it.

### DIFF
--- a/docs/src/tutorials/AtomsBase_interface.md
+++ b/docs/src/tutorials/AtomsBase_interface.md
@@ -67,7 +67,7 @@ ACEpotentials.linear_errors(data_train, model; weights=weights);
 
 In AtomsBase you have system wide features accesses with `haskey(some_key)` or atom features accesses with `hasatomkey(some_key)`.
 Energy (scalar) and virial (matrix) are expected to be system features while force (vector) should be a atom feature.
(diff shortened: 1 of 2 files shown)


cortner:
I think as we switch over to full AtomsBase integration we may choose to enforce units. I'm generally against it, but here it seems actually sensible. To be discussed elsewhere. 

But should this be merged into v0.6.x instead of main?

tjjarvinen:
It should be in both.

The issue with units is that ExtXYZ can not save units. So, there is no reliable way to load them from a file. We would need to think how to deal with this too.

cortner:
> The issue with units is that ExtXYZ can not save units. So, there is no reliable way to load them from a file. We would need to think how to deal with this too.

But the user who is loading the data would need to know what the units are. We could enforce that the user tells ACEpotentials what units to use. We could also use eV, A as defaults. I'm don't have an opinion yet. It will emerge over the next few weeks/months.

cortner:
> It should be in both.

ok for me. How do you handle this? merge first into v0.6.x and then merge that into main? Or merge into main and then backport?

tjjarvinen:
I posted anothen PR to v0.6.x branch #199 . That should make it easy.

cortner:
can I merge both?

tjjarvinen:
yes

PR #200, opened 2024-06-01
Passes E0 and rin0cuts correctly

CheukHinHoJerry:
This PR is to pass E0s and rin0cuts to the new ace model argument correctly. Two questions:

- What format do we want to support for passing E0s to `ace_model`? I would imaging sticking to the old format would be the best for compatibility.
- Are there use cases where the pair basis can have different cut off from the many body basis?

### DIFF
--- a/test/models/test_calculator.jl
+++ b/test/models/test_calculator.jl
@@ -25,8 +25,9 @@ rng = Random.MersenneTwister(1234)
 elements = (:Si, :O)
 level = M.TotalDegree()
 max_level = 15
-order = 3 
-E0s = (-158.54496821, -2042.0330099956639)
+order = 3
+E0s = Dict( :Si => -158.54496821u"eV", 
+            :O => -2042.0330099956639u"eV")
 NZ = length(elements)
 
 function make_rin0cut(zi, zj) 
@@ -47,7 +48,7 @@ model = M.ace_model(; elements = elements, order = order, Ytype = :solid,
 
 ps, st = LuxCore.setup(rng, model)
 
-calc = M.ACEPotential(model, ps, st)
+calc = M.ACEPotential(model, ps, st)   
 
 ##
 
@@ -63,7 +64,7 @@ for ntest = 1:20
    n_O = count(x -> x == 8, AtomsBase.atomic_number(at))
    nlist = PairList(at, M.cutoff_radius(calc))
    efv = M.energy_forces_virial(at, calc, ps_zero, st)
-   print_tf(@test abs(ustrip(efv.energy) - E0s[1] * n_Si - E0s[2] * n_O) < 1e-10)
+   print_tf(@test ustrip(abs(efv.energy - E0s[:Si] * n_Si - E0s[:O] * n_O)) < 1e-10)
 end
 
 println()
--- a/test/models/test_calculator.jl
+++ b/test/models/test_calculator.jl
@@ -26,11 +26,24 @@ elements = (:Si, :O)
 level = M.TotalDegree()
 max_level = 15
 order = 3 
+E0s = (-158.54496821, -2042.0330099956639)
+NZ = length(elements)
+
+function make_rin0cut(zi, zj) 
+   r0 = ACE1x.get_r0(zi, zj)
+   return (rin = 0.0, r0 = r0, rcut = 6.5)
+end
+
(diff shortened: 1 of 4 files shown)


cortner:
> What format do we want to support for passing E0s to `ace_model`? I would imaging sticking to the old format would be the best for compatibility.

Yes, I agree. The more we can keep the interface compatible with the old one the better. 

> Are there use cases where the pair basis can have different cut off from the many body basis?

Yes. But I don't think this is urgent. One of the next steps in the near future will be to rewrite the new ACE models as a stack of calculators, (1) a reference model, e.g. OneBody but maybe more - could include ZBL; (2) an arbitrary pair potentia; (3) a many-body potential. But could also be several many-body potentials ... I suggest to postpone changes to the pair potential cutoff until then.

test/models/test_calculator.jl
```diff
@@ -26,18 +26,46 @@ elements = (:Si, :O)
 level = M.TotalDegree()
 max_level = 15
 order = 3 
+E0s = (-158.54496821, -2042.0330099956639)
```
cortner: This is really cryptic. Can you please redefine as a Dictionary and then change internally whatever is needed to convert that Dictionary to a suitable E0 calculator during the evaluation?

test/models/test_calculator.jl
```diff
@@ -26,18 +26,46 @@ elements = (:Si, :O)
...
+NZ = length(elements)
+
+function make_rin0cut(zi, zj) 
+   r0 = ACE1x.get_r0(zi, zj)
+   return (rin = 0.0, r0 = r0, rcut = 6.5)
+end
+
+rin0cuts = SMatrix{NZ, NZ}([make_rin0cut(zi, zj) for zi in elements, zj in elements])
```
cortner, line 38: We should provide better convenience functions for this. But that can be a separate PR. Maybe file an issue and we can discuss what it should look like...

test/models/test_calculator.jl
```diff
@@ -26,18 +26,46 @@ elements = (:Si, :O)
...
 
 ##
 
+@info("Testing correctness of E0s")
+ps_vec, _restruct = destructure(ps)
+ps_zero = _restruct(zero(ps_vec))
+
+for ntest = 1:20
+   local Rs, Zs, z0, at, efv 
+   at = AB.rattle!(AB.bulk(:Si, cubic=true) * 2, 0.1)
```
cortner: we should improve this test by also "rattling" the elements. I'll try to make a  PR to AtomsBuilder today and we can then update this test before merging.

cortner (changes requested):
Thank you. looks great with just a little bit or minor cleanup needed.

test/models/test_calculator.jl
cortner: P.S.: this is now running through the general registry. I'll ping you again when it is ready.

cortner: Now registered in AtomsBuilder 0.0.4

cortner: I adjusted this test.

test/models/test_calculator.jl
cortner: I also think we are moving to enforcing units right now. So really it should read 
```julia
E0s = Dict( :Si => -158.54496821u"eV", 
             :O => -2042.0330099956639u"eV")
```

CheukHinHoJerry: The output of `model` (not calculator) that depends on `E0s` does not have units though. How about this:

- ustrip `E0s` when the model is built, and put it inside the `meta`
- In the future when the user can change the `energy_unit` when creating potential with `M.ACEPotential`, assert that the unit of the potential is the same as `E0s`. If not, pop a warning and convert the `E0s` units to the same as the calculator.

CheukHinHoJerry: I also put E0s (with units) into `model.mata` since the evaluation of `model.E0` itself does not involves units. I don't want `meta` or `ustrip` everywhere or change the structure too much. Will saving value of `E0s` 2 times be a problem?

cortner (approved):

test/models/test_calculator.jl
cortner: for now this is fine. The ustrip too. We need to review the whole interface once the  AtomsCalculators revision is finished and the AtomsCalculatorsUtilities is ready.

PR #203, opened 2024-06-02
Fix rSH for ACEModel

CheukHinHoJerry:
This PR aims at fixing a small bug for using rSH.

### DIFF
--- a/src/models/ace.jl
+++ b/src/models/ace.jl
@@ -242,8 +242,8 @@ end
 import Zygote
 
 # these _getlmax and _length should be moved into SpheriCart 
-_getlmax(ybasis::SolidHarmonics{L}) where {L} = L 
-_length(ybasis::SolidHarmonics) = SpheriCart.sizeY(_getlmax(ybasis))
+_getlmax(ybasis::Union{SolidHarmonics{L}, SphericalHarmonics{L}}) where {L} = L 
+_length(ybasis::Union{SolidHarmonics, SphericalHarmonics}) = SpheriCart.sizeY(_getlmax(ybasis))
 
 function evaluate(model::ACEModel, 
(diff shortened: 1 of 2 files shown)


cortner:
I only see SH fix, but not the alternative intialization. Should we restrict the PR to the SH fix?

CheukHinHoJerry:
Sure - let's do the rSH fix first.

cortner:
Could we think of a simple test that checks explicitly that the correct Y basis was selected?

CheukHinHoJerry:
How about picking `norm(Rs)` goes to `0`?

CheukHinHoJerry:
- checking relative scaling for random r
- Setting parameters to 1 and 0 and look at the many body basis on `Rs` with `length(Rs)=1`?

cortner:
Produce two bases and compare their behaviour in that limit? That sound good. Would you mind adding this?

cortner:
You don't even have to do r to 0. just check the relative scaling for random r.

CheukHinHoJerry:
I was looking for the `BB_spec` but couldn't find it elsewhere. My memory is that this is something that we should get from `_rpi_A2B_matrix`?

cortner (approved):

CheukHinHoJerry:
Ah okay - over complicate things on my end. Thank you for fixing this.

cortner:
@CheukHinHoJerry -- please confirm that my change to the spher vs solid test is ok. What you had was much more than I had intended and just got a bit too complicated for my taste. If you think there is something crucial missing from my simplified test let me know and we can bring yours back. 

If you are ok with this, then I'll merge.

CheukHinHoJerry:
Looks good. Just to add I feel a bit uncomfortable working without BBspec now. That requires `(nn,ll)` from generating the `A2B` map. Shall we add that feature?

Issue #204, opened 2024-06-03
Bring BBspec back


cortner:
cc @CheukHinHoJerry

CheukHinHoJerry:
This probably best done with a PR at `EQM` so that the `A2B` map gives an option to return the `(nn,ll)`? I can still track that via the `SparseTensor` and `AAspec` though if we don't want to change that.

cortner:
Not necessarily. I think we can just read it off `A2Bmap`. Maybe I can try tonight.

CheukHinHoJerry:
Right - I just realized that should be easier. I am halfway so maybe I can take care of that? Will finish it off tonight.

CheukHinHoJerry:
Done by #205

PR #205, opened 2024-06-03
Add functionality for BBspec `get_nl`

CheukHinHoJerry:
ref : #204

### DIFF
--- a/src/models/sparse.jl
+++ b/src/models/sparse.jl
@@ -58,9 +58,15 @@ function pullback_evaluate(∂B, tensor::SparseEquivTensor{T}, Rnl, Ylm,
 end
 
 
-_nl(bb) = [(n = b.n, l = b.l) for b in bb]
+"""
+Get the specification of the BBbasis as a list (`Vector`) of vectors of `@NamedTuple{n::Int, l::Int}`.
 
-function get_nl(tensor::SparseEquivTensor{T}) where {T}
+### Parameters 
+
+* `tensor` : a SparseEquivTensor, possibly from ACEModel
+"""
+function get_nnll_spec(tensor::SparseEquivTensor{T}) where {T}
+   _nl(bb) = [(n = b.n, l = b.l) for b in bb]
    # assume the new ACE model NEVER has the z channel
    spec = tensor.aabasis.meta["AA_spec"]
    nBB = size(tensor.A2Bmap, 1)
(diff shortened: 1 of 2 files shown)


CheukHinHoJerry:
I think we don't really need to store the `BBspec` in the model, given that we didn't do that (as far as I know) for previous `acemodel` too.

src/models/sparse.jl
```diff
@@ -57,3 +57,21 @@ function pullback_evaluate(∂B, tensor::SparseEquivTensor{T}, Rnl, Ylm,
    return ∂Rnl, ∂Ylm
 end
 
+
+_nl(bb) = [(n = b.n, l = b.l) for b in bb]
```
cortner: maybe move this *inside* `get_nl`

src/models/sparse.jl
```diff
@@ -57,3 +57,21 @@ function pullback_evaluate(∂B, tensor::SparseEquivTensor{T}, Rnl, Ylm,
...
+
+function get_nl(tensor::SparseEquivTensor{T}) where {T}
+   # assume the new ACE model NEVER has the z channel
```
cortner, line 70: yes, I think we want to guarantee *HERE* at least that there are only `(n, l, m)` channels. If that ever changes then we will of course have to revisit.

src/models/sparse.jl
```diff
@@ -57,3 +57,21 @@ function pullback_evaluate(∂B, tensor::SparseEquivTensor{T}, Rnl, Ylm,
    return ∂Rnl, ∂Ylm
 end
 
+
+_nl(bb) = [(n = b.n, l = b.l) for b in bb]
+
+function get_nl(tensor::SparseEquivTensor{T}) where {T}
```
cortner: can we call this `get_nnll_spec` please? Can can you add a short inline documentation also to explain what this returns?

cortner (changes requested):
Thank you. This looks good, can you please just take care of the minor issues I pointed out? 

I'm not sure how to write tests for this, so let's just not worry about it, cross our fingers and hope for the best?!? If anybody has an idea then we should of course add a unit test here.

src/models/sparse.jl
CheukHinHoJerry: The reason I call this `get_nl` is that `ACE1` and `ACEpotentials` calls that too. But this is a new model type so we can probably not worry about it?

cortner (approved):

src/models/sparse.jl
cortner: I see - I prefer we don't tie ourselves to bad decisions made in the past :)

cortner:
happy to merge?

CheukHinHoJerry:
Yes please.

PR #207, opened 2024-07-02
Update to P4ML v0.3.x

cortner:
P4ML had a serious cleanup with v0.3.x, including removing ObjectPools and tidying up the interface. This PR will update ACEpotentials to the latest P4ML, and then try to replace as many allocating layers with suitable usage of Bumper.


cortner:
- [x] The status of this PR is that all tests pass now; except for two that have been failing for a while.
- [x] Many "layers" have been updated to be non-allocating, but not all; this can be done in future PRs.
- [x] TODO: check the example script still runs

PR #209, opened 2024-07-20
Weight initializations to mimic linear ACE1

cortner:
This PR is to implement weight initializations (mostly for the radial basis) to be able to reproduce ACE1 results within ACE2. 

~At the moment, this only supports single species. I will implement the multi-species setting asap.~


cortner:
@CheukHinHoJerry  -- this is basically ready to look at but I can't yet reproduce ACE1 accuracy. Something we need to look at carefully. But maybe it is enough to start working on the parameter estimation.

CheukHinHoJerry:
Thank you. I will take a look before Friday. (Please feel free to merge if you think this is ready.)

cortner:
Just a brief update on what is going on with this PR: I've written a separate script that looks into the details of the radial and pair bases. 
- [x] It appears that the radial basis looks ok but fairly different from that of ACE1. I would like to match the two more closely before completing this PR. 
- [x] The pair basis is complete rubbish. Something has gone very wrong. **EDIT:** this is now fixed.

cortner:
Despite having a seemingly very sensible basis in ACE2, I am still getting terrible fit accuracy compared with ACE1. This is very unexpected for me. The only idea I have so far is to work towards exactly matching the radial and pair bases and then to confirm that ACE1 and ACE2 span the same or a very similar space.

CheukHinHoJerry:
Previously I had much better experience using spherical harmonics instead of solid, could that help?

cortner:
Very good point. I'll look at that next. Thank you!

cortner:
Switching solid to spherical harmonics did not fix the accuracy discrepancy.  The following is Zuo / Si. For other elements the discrepancy is much worse.
```
Model  |     E    |    F  
 ACE1  | 1.04e-02  |  3.12e-01  
 ACE2  | 1.60e-01  |  5.08e-01  
```

cortner:
Just to document where we are. The pair and radial bases now match perfectly, but even 2-correlations with non-trivial angular dependence do not. This is despite the fact that the coupling coefficients look correct.

cortner:
@CheukHinHoJerry  -- this looks ready to experiment with. I'd be grateful if you briefly skim the following scripts in `docs/src/newkernels` : 
- test_matching_bases.jl : confirm that an ACE1 and new basis span the same space
- linear_match.jl : linear regression with ACE and new basis, closely matching errors
- linear_nonlinear.jl : an example script doing linear regression followed by nonlinear 

Unless you have any concerns about those tests, I will merge this PR and we could then start a new PR where we provide a full compatibility layer to ACE1.

cortner:
Thanks jerry. I'll merge this so it is in the main branch and we can start creating separate PRs to work from here.

cortner:
hm - this was closed accidentally instead of merged.... weird.

Issue #208, opened 2024-07-04
Explosion of energy per atom on ideal lattice sites with increasing supercell size


ConnorSA:
Hey all,

I noticed some strange behaviour when using ACE potentials I developed and wondered if this could be a potential bug. In particular, I found that the energy per atom would explode sometimes when increasing super cell size when atoms sat on the lattice sites. Here is a quick python snippet of the test:

```
atoms=ase.io.read('path/to/Ti_bcc_primitive.castep')
ace_potential="path/to/potential.json"
ACE = pyjulip.ACE1(ace_potential)
for i in range(1,10):
    a=atoms.copy()
    a=a*[i,i,i]
    a.calc=ACE
    print(a.get_potential_energy()/len(a))
```
which would output the following atomic energies:
```
-1593.736219589828
-1593.7362195898277
-1593.7362195898286
-1593.7362195898268
-1593.7362195898286
-1477.8662321399227
7594.737651725763
33464.98197817054
433759.4730885219
```
If I rattle the atomic positions by a tiny amount (0.00001 \AA), then I recover something more in line with what I expected in terms of atomic energy per atom at large super cell sizes (for reference this is -1593.72718054 eV from DFT). I think something strange is happening numerically when atoms happen to sit in particular configurations, wondered if this has been seen before and if it's reproducible? I'm not sure how consequential this might be, but wanted to throw this out there :)

I have managed to get this effect with different ACE potential complexities, but it seems to be inconsistently appearing in particular crystal symmetries (I haven't had this effect on an orthorhombic box yet).

As far as ACE parameters go, I've managed to get this behaviour with different orders and degrees in the model complexity. In this case this is with a order=3, total_degree=15, and cutoff of 6.0 \AA.

julia:
ACEpotentials v0.6.2

Many thanks,

Connor
University of Warwick

cortner:
Thanks for reporting this. Feels like some array hasn't been initialized as expected. 

- Are you in a position to produce a pure Julia test? 
- If not, can you comment on whether this happens with ANY ACE1 potential or just the specific one you used here? 
- If only this specific potential, then can you please share the JSON file? 

EDIT: apologies I didn't read carefully enough. It sounds like I would be able to recreate this with randomly generated ACE potentials?

cortner:
On a first attempt I cannot reproduce this, here is my script: 
```julia
using ACEpotentials 

model = acemodel(elements = [:Ti, ],
					  order = 3,
[...]

PR #213, opened 2024-07-30
Add test for #208

cortner:
This PR adds a simple test that checks for #208. Once we confirm failure on 1.9 but pass on 1.10 we can merge, add new version bounds and move on. 

.... for simplicity I will rebase onto v0.6.x and then forward-port to main


cortner:
TODO: fix test tolerance, version bounds.

PR #214, opened 2024-07-30
Add test for #208 to v0.6

cortner:
Adds a test for the bug discussed in #208  to branch v0.6.x


cortner:
CI shows that on 1.9 the test fails while on 1.10 the test passes. This was reproduced on two additional Linux machines. Next steps: 
- remove Julia 1.9 as an option
- merge
- bump version. Technically this is breaking but we will keep 0.6.x since this branch is about to be retired.

PR #219, opened 2024-08-07
Revisiting Smoothness Priors

cortner:
So far this PR gives just a simple re-implementation of the old ACE1x smoothness priors. But I would like to rethink how the priors interact with basis construction and rescaling of the radial basis. Maybe this needs to become a separate PR. We will see. 

CC @CheukHinHoJerry @iamveronika


cortner:
this makes smoothness priors compatible with usage as it was in ACE1. I suggest we merge this and integrate it into Veronika's work. 

As I suggest above, the integration of smoothness priors into the basis itself can become a separate PR. I don't want to rush this and write it down very carefully with all details written out and discussed and another discussion how this should integrate with nonlinear models.

cortner:
@CheukHinHoJerry  - if you have no concer or objection I will go ahead and merge this?

CheukHinHoJerry:
Yes please. No concern at all from my end.

Issue #226, opened 2024-08-22
Getting the CI back on track


wcwitt:
I'm unhappy that the CI is failiing and would like to get it working again.

One issue seems to be this line in the `Project.toml`, which (if I understand correctly) restricts the `julia` version such that the tests fail on nightly (currently 1.12.0-DEV.1081).
```
julia = "~1.10.0"
```
What is the appropriate fix here? Does anyone mind if I simply remove the tilde?

wcwitt:
Will be resolved by #229.

PR #227, opened 2024-08-22
WIP: Towards working CI.

wcwitt:
See #226. I'm creating this PR to trigger the CI. I won't merge without confirmation from someone.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -62,7 +62,7 @@ PrettyTables = "1.3, 2.0"
 Reexport = "1"
 StaticArrays = "1"
 YAML = "0.4"
-julia = "1.10.0"
+julia = "~1.10.0"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
--- a/Project.toml
+++ b/Project.toml
@@ -62,7 +62,7 @@ PrettyTables = "1.3, 2.0"
 Reexport = "1"
 StaticArrays = "1"
 YAML = "0.4"
-julia = "~1.10.0"
+julia = "1.10.0"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -15,5 +15,5 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-ACEpotentials = "0.6"
+ACEpotentials = "0.8"
 Literate = "2.13.2,3"
--- a/docs/src/tutorials/smoothness_priors.jl
+++ b/docs/src/tutorials/smoothness_priors.jl
@@ -32,9 +32,9 @@ A, Y, W = ACEfit.assemble(data, model.basis);
 
 # In the following we demonstrate the usage of algebraic and gaussian priors. The choices for `σl, σn` made here may seem "magical", but there is a good justification and we plan to automate this in future releases. 
 
-Pa2 = algebraic_smoothness_prior(model.basis; p=2)
-Pa4 = algebraic_smoothness_prior(model.basis; p=4)
-Pg  = gaussian_smoothness_prior( model.basis, σl = (2/r_nn)^2, σn = (0.5/r_nn)^2);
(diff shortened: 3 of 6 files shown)


wcwitt:
IMPORTANT. I've disabled the experimental tutorial because I don't know how to fix it myself. Need to revive it before this is merged.

Otherwise, I am making progress. @tjjarvinen do you understand this documentation error? I think it's this but not certain.
> These are docstrings in the checked modules (configured with the modules keyword)
that are not included in canonical @docs or @autodocs blocks.

cortner:
Doesn't this just mean there are new functions defined with doc strings and they should be listed in the docs. I think I had this happen somewhere else? (I might misremember...)

wcwitt:
It does seem that way, but we definitely don't mention all functions by name in the docs; instead there is some automatic feature that finds them. Why can't it find these? I need to remind myself of the details.

Project.toml
```diff
@@ -62,7 +62,7 @@ PrettyTables = "1.3, 2.0"
 Reexport = "1"
 StaticArrays = "1"
 YAML = "0.4"
-julia = "~1.10.0"
```
cortner: I thought the ~ means that we restrict to 1.10.x and don't allow 1.11. This was deliberate. Did I misunderstand something?

docs/Project.toml
```diff
@@ -15,5 +15,5 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-ACEpotentials = "0.6"
+ACEpotentials = "0.8"
```
cortner, line 18: Should there even be a compact entry? Shouldn't the docs build just dev the "current" ACEpotentials?

docs/src/tutorials/smoothness_priors.jl
```diff
@@ -32,9 +32,9 @@ A, Y, W = ACEfit.assemble(data, model.basis);
 
 # In the following we demonstrate the usage of algebraic and gaussian priors. The choices for `σl, σn` made here may seem "magical", but there is a good justification and we plan to automate this in future releases. 
 
-Pa2 = algebraic_smoothness_prior(model.basis; p=2)
-Pa4 = algebraic_smoothness_prior(model.basis; p=4)
-Pg  = gaussian_smoothness_prior( model.basis, σl = (2/r_nn)^2, σn = (0.5/r_nn)^2);
+Pa2 = ACEpotentials.algebraic_smoothness_prior(model.basis; p=2)
```
cortner, line 35: I think this is wrong. It should be `ACE1x.algebratic_smoothness_prior` etc .... But we are almost ready to get rid of ACE1 and ACE1x entirely so I'm not sure it's worth working on this right now.

Project.toml
wcwitt: Everything in this PR is up for discussion, no question. But this CI failure on nightly (which is running 1.12.0-DEV) was caused by the `~`, as best I can tell, and I would prefer to always have the CI passing on `main`.

https://github.com/ACEsuit/ACEpotentials.jl/actions/runs/10500352832/job/29088598779

cortner: Then turn off nightly

docs/src/tutorials/smoothness_priors.jl
wcwitt: Not sure, I chose `ACEpotentials` because of this https://github.com/ACEsuit/ACEpotentials.jl/blob/4a09741259e370b97a83f18bacc47877a6d0eae2/src/models/smoothness_priors.jl#L90

I don't mind changing to `ACE1x` if that's better, but at minimum the change I made allows that part of the docs to build successfully, where it failed before.

I don't care about this example specifically, but I want to fix the unimportant CI failures so that more impactful things don't get lost.

Project.toml
wcwitt: I can live with that https://github.com/ACEsuit/ACEpotentials.jl/pull/227/commits/0c1b7cccdeca2e52430560bafef63036ce1a4146

docs/Project.toml
cortner: This is the line I get when building docs via CI: 
```
  [3b96b61c] + ACEpotentials v0.8.0-dev `~/work/ACEpotentials.jl/ACEpotentials.jl`
```
So the compat entry in the docs/Project folder is really not needed (and shouldn't be there)

wcwitt:
Closing, replaced by #229.

PR #232, opened 2024-08-27
WIP:  Add New Kernels to All Tests

cortner:
- [x] Update the interface to AC 0.2 and ACU 0.1 
- [x] Make all current tests pass for both ACE1 type models and for the new kernels.
- [x] Remove ACEmd dependency 

In addition 
- I will slim down the tests a little bit. It seems that testing different solvers can be done within ACEfit; here it should be sufficient to test just a select few. That way there is less to maintain. --- I did this a bit, but more cleanup is needed.
- I'll try to recover the target accuracies somehow.  --- here I failed. Not sure what changed. 
- Prepare the package for re-introducing the JSON interface. --- this seems ready


cortner:
This requires ACEfit.jl#79 to be merged first (cc @wcwitt )

PR #233, opened 2024-08-28
WIP: Trace

cortner:
This PR will implement the symmetric (and time permitting also the non-symmetric) trace models. 

NB It branches off from #232 , so that will have to be merged first.


cortner:
CC @wcwitt  I didn't manage to finish tonight, but I may be able to finish it by end the of the week. Of course, getting a crappy first sketch is not hard. Making it work on real problems and real datasets will be the challenge...

cortner:
First draft ready, and could actually already be tried out. non-symmetric version still to be added.

cortner:
all this work has moved to `EquivariantTensors` so I'm closing this but keeping the branch to maybe ctransfer the code later.

PR #238, opened 2024-09-01
Retire ACE1

cortner:
- [x] remove all dependencies on ACE1, ACE1x
- [x] rewrite all tests to not require ACE1, ACE1x
- [x] rewrite the documentation, possibly slimmed down initially.

Further Notes
- I'm removing the fast experimental kernels (UltraFastACE) for now
- Writing and reading of models will go into a separate PR
- Export to LAMMPS will not be immediately updated.


cortner:
close to ready to merging. But needs PRs in ExtXYZ and AtomsBase merged first.

cortner:
@wcwitt  -- the tests are failing here, but in fact that's only because they rely on two PRs that haven't been merged yet, one in AtomsBase and one in ExtXYZ. 

Tomorrow I want to add merge the JSON interface, convert the tutorials to using that interface, then add a test suite that confirms that the error tables from the ACEpotentials paper are reproduced. If this can all be done, then I think we may be able to do the tutorial on 0.8.

PR #243, opened 2024-09-10
Tweaks following a full pass through the documentation

wcwitt:
Overall, everything in the `dev` docs is pretty clean. A few small things:

* The weights should be squared here, right?
image

* There is one TODO in the smoothness prior tutorial
image


cortner:
> The weights should be squared here, right?

yes. Thanks for spotting this.

wcwitt:
Just pushed a fix for the loss function - let's see how it renders.

wcwitt:
The latest, from https://acesuit.github.io/ACEpotentials.jl/previews/PR243/literate_tutorials/basic_julia_workflow/

image

cortner:
This looks like I should merge, yes?

wcwitt:
>There is one TODO in the smoothness prior tutorial

This is the only lingering item, but not essential to fix it now. Fine to merge if you like.

cortner:
I'll merge this and let's file an issue to integrate smoothness priors into runfits as well as update the tutorials.

PR #244, opened 2024-09-10
Starting a test suite

cortner:
Starting a test suite, ideally bring it to parity with ACEpotentials paper last year. So far tests don't run because of memory problems. To be investigated.

### DIFF
--- a/examples/zuobench/error_table.jl
+++ b/examples/zuobench/error_table.jl
@@ -1,13 +1,15 @@
+# This script reproduces the error table from the 2023/24 ACEpotentials 
+# paper, but with the new version 0.8 ACE models.
+
 using Distributed 
 addprocs(10, exeflags="--project=$(Base.active_project())")
 @everywhere using ACEpotentials, PrettyTables
 
 # the dataset is provided via ACE1pack artifacts as a convenient benchmarkset
 # the following chemical symbols are available:
-# syms = [:Ni, :Cu, :Li, :Mo, :Si, :Ge]
-syms = [:Ni, :Cu, ] 
+syms = [:Ni, :Cu, :Li, :Mo, :Si, :Ge]
 
-totaldegree_tiny = [ 18, 14, 10 ]   # very small model: ~ 100  basis functions
+totaldegree_tiny = [ 16, 12, 8 ]   # very small model: ~ 120  basis functions
 totaldegree_sm = [ 20, 16, 12 ]   # small model: ~ 300  basis functions
 totaldegree_lge = [ 25, 21, 17 ]  # large model: ~ 1000 basis functions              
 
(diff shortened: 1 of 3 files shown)


cortner:
The memory problems are fixed by #246 . The first zuo example script now runs and produces ok results. 

I'm now trying to add an ASP example on one of the Zuo datasets, but the solver returns only the last result and not the path. Probably a bug (might have been introduced by me).  @tinatorabi  --  would you mind looking into this?  If I figure something out myself, I'll let you know immediately.

cortner:
```
Energy Error
┌────┬─────────┬──────────┬──────┬──────┐
│    │ ACE(sm) │ ACE(lge) │  GAP │  MTP │
├────┼─────────┼──────────┼──────┼──────┤
│ Ni │   0.409 │    0.334 │ 0.42 │ 0.48 │
│ Cu │   0.291 │     0.22 │ 0.46 │ 0.41 │
│ Li │   0.233 │    0.166 │ 0.49 │ 0.49 │
│ Mo │   2.386 │     2.88 │ 2.24 │ 2.83 │
│ Si │   3.522 │    2.305 │ 2.91 │ 2.21 │
│ Ge │   2.669 │    2.192 │ 2.06 │ 1.79 │
└────┴─────────┴──────────┴──────┴──────┘
Force Error
┌────┬─────────┬──────────┬──────┬──────┐
│    │ ACE(sm) │ ACE(lge) │  GAP │  MTP │
├────┼─────────┼──────────┼──────┼──────┤
│ Ni │   0.018 │    0.015 │ 0.02 │ 0.01 │
│ Cu │   0.007 │    0.005 │ 0.01 │ 0.01 │
│ Li │   0.006 │    0.005 │ 0.01 │ 0.01 │
│ Mo │   0.122 │    0.096 │ 0.09 │ 0.09 │
│ Si │   0.086 │    0.065 │ 0.07 │ 0.06 │
│ Ge │   0.063 │    0.051 │ 0.05 │ 0.05 │
└────┴─────────┴──────────┴──────┴──────┘
```
Looking ok so far. I think this makes me comfortable to merge this PR and fix the ASP test asap.

cortner:
Screenshot 2024-09-11 at 13 41 22

For comparison this is the result from the paper. Essentially the same.

tinatorabi:
@cortner I tried this and once you have the results via `results= ACEfit.solve(solver, AW, YW, AWv, YWv)`, `results["C"]` gives you the best model parameters on the validation set based on the chosen selection mode and to get the parameters for the full path you'd have to do something like `[p[:solution] for p in results["path"]]`. Do we need this returned as well?

cortner:
I missed that there was a kwarg to get the path. I've now removed that and ALWAYS specify this.

Issue #242, opened 2024-09-10
Shell script should offer specification of output filename


cortner:
Currently the shell script works as follows
```shell
julia --project=. runfit.jl -p model_spec.json
```
with the output file specified in `model_spec.json`.  There should be another argument that overrides the output file spec, e.g. 
```shell
julia --project=. runfit.jl -p model_spec.json -o result.json
```

cortner:
@iamveronika  -- could you make a PR that adresses this?

cortner:
maybe coordinate with @CheukHinHoJerry  - to see how he envisions using the runfit script.

CheukHinHoJerry:
`julia --project=. runfit.jl -p model_spec.json -o result_folder` seems to be the best for me, and then maybe primarily in the `output` section of `model_spec.json` we can support a few easy tests
```
  "output": {
    "model": "opted_model",
    "dimer": false,
    "error_table": true,
    "scatter_error": false
  }
```
and we output everything into `result_folder` together with the data needed to reproduce those plots in a `result_folder/result.json` file.

cortner:
and maybe the default model filename can just be `model.json`, unless otherwise specified in the model specification dictionary?

cortner:
I think something like this will be great.

cortner:
and can we take a default output folder to be `result_model_spec` where `model_spec` is the input filename? (of `-o` is not specified)

CheukHinHoJerry:
yes same thought, i will create a PR after finishing this items.

cortner:
great, please make sure to merge latest `main` into your branch.

cortner:
Now addressed by #254

PR #254, opened 2024-09-12
WIP: add shell output file and simple analysis

CheukHinHoJerry:
Ref : #242 

Things added:

1. We can do 

`julia --project=. runfit.jl -p model_spec.json -o result_folder`

now. All the results are stored in the result_folder

2. Analysis such as dimer and scatter plots for E, F and V are added. User can specify `make_plots = true` in the input dictionary to do the plots. And can turn on/off different analysis that is saved together with the model in `result_folder/results.json`.
image

Example scatter plot with a tiny model on an AgPd dataset

scatter


cortner:
Looks very nice, thank you. ready to merge from your end?

CheukHinHoJerry:
Sorry for late reply. Yes I think it is ready. We can improve on the plots later.

PR #264, opened 2024-09-15
Committees 

cortner:
this is a poor-man's prototype implementation. Can be done much more efficiently. Just to get the tutorial running and get closer to feature parity.


cortner:
this isn't really complete, it is missing integrating the committee into acefit; documentation; saving of committees, and likely some other things. the plan is to merge it into main, then into the tutorial PR. While finishing the tutorial we can wrap up some of those points.

PR #277, opened 2024-10-22
Fix tutorial path string escaping

nhz2:
Using `repr` here is more reliable if the path contains special characters.

This fixes `copy_tutorial` on Windows.


cortner:
wonderful, thanks for the PR

cortner:
I'll fix another bug and the register a new version

PR #303, opened 2025-11-14
WIP EquivariantTensors.jl  / Julia 1.12 port

jameskermode:
This is an initial port of ACEpotentials.jl from a backend of EquivarientModels.jl to EquivariantTensors.jl. At the same time I have added support for Julia 1.12 and Lux 1.x.

After an initial planning and iteration phase, the port was largely completed by Claude Code so will need careful review and iteration before it can be merged. For sure there will be some naive/non-idiomatic usages we will want to clean up.

Most tests are passing, but note that I had to increase the tolerance on the RMSEs in test_silicon.jl. I also had to disable the test_bugs.jl test since that is failing on Julia 1.12. It still passes on Julia 1.11; I think the failures are due to changes in dictionary sort order so we should move to OrderedDict throughout to fix this.

I also still need to understand why the same (order, totaldegree) parameters lead to different final basis set sizes on ET and EM. For example (3, 10) on EM gives 120 basis functions, but 77 with the new ET port.

I first tested on energies and once those matched well I renabled forces and virials. I removed the custom adjoint rules and instead used ForwardDiff to compute forces (after checking that this was more efficient than Zygote). Performance is ca. 20% slower for assembly than the previous custom adjoints, but overall performance is comparable because of the reduced basis set sizes.


jameskermode:
Oh, and I also disabled the fast evaluator for now as that would need significant work to update to ET.

cortner:
Thank you James for taking this on. I'll do my best to take a close look by Tuesday. 

> I also still need to understand why the same (order, totaldegree) parameters lead to different final basis set sizes on ET and EM. For example (3, 10) on EM gives 120 basis functions, but 77 with the new ET port.

I'll try and spend a bit of time on this.

> Oh, and I also disabled the fast evaluator for now as that would need significant work to update to ET

I think that remained experimental and can be permanently retired. Some of the ideas in that package can be moved into ET.

cortner:
I'm just merging this into a local branch now if that's ok, then I can start taking a look and making changes conveniently. Feel free to just directly work in `dev-0.10` or make new PRs to `dev-0.10` either from your fork or from local branches, whatever is most convenient for you.

PR #304, opened 2025-11-15
EquivariantTensors + J1.12 Port

cortner:
MY TODO LIST 

- [ ] replace all model evaluation code with ET
- [x] understand (and fix?) the basis size issue

---

@jameskermode  --  Feel free to just directly work in `dev-0.10` or make new PRs to `dev-0.10` either from your fork or from local branches, whatever is most convenient for you.

Thank you James for taking this on. I'll do my best to take a close look by Tuesday. 

> I also still need to understand why the same (order, totaldegree) parameters lead to different final basis set sizes on ET and EM. For example (3, 10) on EM gives 120 basis functions, but 77 with the new ET port.

I'll try and spend a bit of time on this.

> Oh, and I also disabled the fast evaluator for now as that would need significant work to update to ET

I think that remained experimental and can be permanently retired. Some of the ideas in that package can be moved into ET.


cortner:
Had a first closer look, the refactoring is fairly superficial right now - the new backends are not really used yet, but this is now the part that I will need to take on I think. (will create new branch for this.)

Because of this, I must admit I am now also a bit confused about the basis size issue. This might take me longer to understand.

jameskermode:
I wondered about that. At least we have an ET and Julia 1.12 compatible framework to build on, and the tests and infrastructure are working.

cortner:
Yes! And if I'm not mistaken this is now compatible with the General registry. So as soon as we are happy with a prototype we could register it.

jameskermode:
I'd be happy to do that once we understand basis set size and RMSE regressions (which are probably related to one another)

cortner:
@jameskermode fairly certain the last commit fixes the basis size bug. Can you re-check tests? I don't remember what if anything you changed from the last minor version? But any changes should now probably be reverted. 

P.S.: if you take a look at the bug it shows you that even Claude isn't quite ready :)

jameskermode:
Thanks! That was indeed an error by Claude, it has been duly disciplined and told to learn from its mistakes. I've reverted the loosening of the RMSE thresholds and confirmed they now pass locally and the basis sizes are as expected. If they pass in CI then we're all good.

jameskermode:
Once you are happy with your prototype in #305, I think Claude could be trusted (with supervison) to migrate more of the main code to use it. See it's updated understanding in CLAUDE.md in 900bfdc if you are interested.

cortner:
We can definitely try that. But I'd like to make that transition only once everything is ready. If we do it too soon then a lot of functionality and performance may be lost.

cortner:
In the meantime, do you think this is ready to merge into `main` and register at `0.10.0`?

jameskermode:
Providing nothing downstream is critically dependent on the fast evaluator, then yes, we can merge.

cortner:
I think the fast evaluator is used in a few cases to compute hessians. Otherwise it is only important for performance. I can probably bring it back in with not too much effort if you think it is warranted. 

The hessians will be provided with the new kernels. The performance will require a fair bit of work to replicate with the new kernels because they are optimized for flexibility not performance.

jameskermode:
@cortner If you are still happy to do this I will go ahead and merge this PR as is, then register at 0.10.0 in General

cortner:
I think that's ok - with a warning in the README that 0.10 drops support for the fast evaluator. (Unless you are asking me to bring that support back in??)

cortner:
It would be good to register something asap so that we are in General and can start registering new versions as they get ready.

jameskermode:
No, not asking you to add the fast evaluator, will add a warning then merge and register.

cortner:
Great, thank you!!

PR #305, opened 2025-11-18
EquivariantTensors Backend

cortner:
dev-0.10 uses ET only superficially, this PR is to integrate ET more deeply, ~remove duplicated functionality~, with the end-goal to enable GPU evaluation.

- [x] complete ET+Lux model, match old-style
- [x] ~forces~ site energy gradients
- [x] ~basis~ site energy basis and site energy gradient basis  
- [x] integrate as alternative model that acts exactly the same way
- [x] full GPU evaluation
- [x] one-body
- [x] pair potential (#316)
- [x] ET splines (#317)
- [x] calculators (#313)
- [x] tests (#313)
- [x] assembly (#313)

Removing duplicated functionality will be done in a future version.


jameskermode:
See #312 which targets this PR and implements most of the remaining steps (but still needs to be cleaned up)

cortner:
UPDATE: I think I now have all functionality that we can expect ET to provide. My plan is to finalize the unit tests and then I think we can make a plan how to proceed. Here is a tentative list: 

## Backend
- splines (still missing)
- gpu evaluation (I broke this again)

## Low model level 
- pair potentials, E0s => I'd like to treat these as stacked models instead of integrated, to be discussed.
- wrap the new ACE model into a calculator that takes care of energy, forces, virial, etc
- provide a wrapper that generates an old ACE model, then converts to new
- extensive correctness tests and unit tests to compare the two models for correctness and performance 

## User-facing 
- Throuhgout 0.10 keep both frameworks alive 
- For 0.11 retire the entirety of ACE 0.8, 0.9 infrastructure which is still surviving in 0.10

## Speculative
- bring back purification

cortner:
- [ ] The dictionary test passes on my laptop so I removed the testfail but it still fails on CI. Will need to bring it back before finishing this PR

cortner:
@jameskermode  --  this now has one-body and pair potentials and is in principle at a place where one could put together a complete ACE model, basis and calculators and start testing backward compatibility. The only technical thing left to do for me is testing the splines. (the functionality is in principle already there...)

jameskermode:
Great, I'll rebase my calculators PR on this and continue

cortner:
I realized I made a mistake in the design of the basis interface. I'm returning the site energy basis but for each center-atom species, the basis occupies the same indices. We need to perform a transformation so that bases for different species occupy separate indices. A question is whether this should be done at the site potential model level or at the calculator level. Maybe it can be done at the level of the site potential wrapper calculator? But it may need some additional logic. we should discuss what a good general-purpose approach is before charging ahead I think.

cortner:
@jameskermode  -  I will merge once tests pass, unless I hear otherwise.

jameskermode:
Go ahead. I suggest we tag this as v0.10.1. LAMMPS export can then come in v0.10.2 in case it causes any problems.

PR #313, opened 2025-12-23
ETACE Calculator Interface and Training Assembly

jameskermode:
## Summary

Adds EquivariantTensors-based calculators for the ETACE backend, enabling:
- AtomsCalculators-compatible energy/forces/virial evaluation
- Linear least squares assembly for model training
- Composable calculator stacking via StackedCalculator
- **GPU acceleration** for energy and forces (requires Polynomials4ML v0.5.8+)

### New Components

**Core Calculators:**
- **ETACEPotential** (type alias): Standalone ACE calculator with full AtomsCalculators interface
- **WrappedSiteCalculator{M, PS, ST}**: Generic wrapper for site energy models implementing the ETACE pattern
- **StackedCalculator**: Combines multiple calculators with @generated loop unrolling

**ET-Native Models:**
- **ETACE**: Many-body ACE basis using EquivariantTensors graph evaluation
- **ETPairModel**: Pair potential model using ET graph evaluation
- **ETOneBody**: One-body reference energies (zero gradients)

**Conversion Functions:**
- `convert2et(model)`: Convert many-body ACE to ETACE
- `convert2et_full(model, ps, st)`: Convert full model (E0 + pair + many-body) to StackedCalculator

### Training Assembly

- `length_basis(calc)` - Total number of linear parameters
- `get_linear_parameters(calc)` / `set_linear_parameters!(calc, θ)` - Parameter management
- `potential_energy_basis(sys, calc)` - Energy design matrix row
- `energy_forces_virial_basis(sys, calc)` - Full EFV design matrix row

### Benchmark Results

**GPU Many-Body Only (ETACE) - Energy:**
| Atoms | Edges | CPU (ms) | GPU (ms) | GPU Speedup |
|-------|-------|----------|----------|-------------|
| 64 | 2146 | 3.38 | 0.54 | **6.3x** |
| 512 | 17176 | 27.77 | 0.66 | **41.9x** |
| 800 | 26868 | 37.12 | 0.78 | **47.6x** |

**GPU Many-Body Only (ETACE) - Forces:**
| Atoms | Edges | CPU (ms) | GPU (ms) | GPU Speedup |
|-------|-------|----------|----------|-------------|
| 64 | 2146 | 46.46 | 14.42 | **3.2x** |
| 512 | 17178 | 104.39 | 15.12 | **6.9x** |
| 800 | 26860 | 289.32 | 16.33 | **17.7x** |

**GPU Full Model (E0 + Pair + Many-Body) - Energy:**
| Atoms | Edges | CPU (ms) | GPU (ms) | GPU Speedup |
|-------|-------|----------|----------|-------------|
| 64 | 2140 | 3.40 | 0.94 | **3.6x** |
| 512 | 17166 | 31.18 | 0.95 | **32.9x** |
| 800 | 26858 | 45.16 | 1.24 | **36.4x** |

**GPU Full Model (E0 + Pair + Many-Body) - Forces:**
| Atoms | Edges | CPU (ms) | GPU (ms) | GPU Speedup |
|-------|-------|----------|----------|-------------|
| 64 | 2134 | 24.05 | 19.34 | **1.2x** |
| 512 | 17178 | ~110 | ~20 | **~5x** |
| 800 | 26860 | ~300 | ~22 | **~14x** |

**CPU Full Model Forces (ETACE vs ACE):**
| Atoms | Edges | ACE CPU (ms) | ETACE CPU (ms) | ETACE Speedup |
|-------|-------|--------------|----------------|---------------|
| 64 | 2146 | 73.6 | 30.5 | **2.4x** |
| 256 | 8596 | 307.7 | 74.4 | **4.1x** |
| 800 | 26886 | 975.0 | 225.6 | **4.3x** |

**Notes:**
- GPU forces now work with Polynomials4ML v0.5.8+ (bug fix in P4ML release Dec 29, 2024)
- GPU shows excellent scaling: larger systems see better speedups
- Full model GPU speedups are lower than many-body only due to graph construction overhead
- CPU forces are 2-4x faster with ETACE due to Zygote AD through ET graph

### Design Note: Basis Index Handling

The current implementation handles species-separated basis indices at the calculator level (in `energy_forces_virial_basis`). Each species gets separate parameter indices via `p = (s - 1) * nbasis + k`. This design choice may need discussion for future GPU training assembly use cases.

### Files Added/Modified

**New files:**
- `src/et_models/et_ace.jl` - ETACE model implementation
- `src/et_models/et_pair.jl` - ETPairModel implementation
- `src/et_models/et_onebody.jl` - ETOneBody implementation
- `src/et_models/et_calculators.jl` - Calculator wrappers and conversion functions
- `src/et_models/stackedcalc.jl` - StackedCalculator with @generated
- `src/et_models/convert.jl` - Model conversion utilities
- `test/etmodels/test_etbackend.jl` - ETACE tests
- `test/etmodels/test_etpair.jl` - ETPairModel tests
- `test/etmodels/test_etonebody.jl` - ETOneBody tests
- `benchmark/gpu_benchmark.jl` - GPU benchmark script

**Modified files:**
- `src/et_models/et_models.jl` - Updated includes and exports
- `Project.toml` - EquivariantTensors compat bumped to 0.4.2
- `test/Project.toml` - EquivariantTensors compat bumped to 0.4.2, added LuxCUDA

## Test plan

- [x] All existing tests pass (945 passed, 1 known Julia 1.12 hash ordering issue)
- [x] New test files cover ETACE, ETPairModel, and ETOneBody
- [x] Finite difference validation for forces and virial
- [x] GPU benchmark confirms significant speedups for energy and forces
- [x] CPU benchmark confirms ETACE force speedups vs classic ACE


cortner:
@jameskermode  -  please move ET compat to 4.3 when you have a moment.

src/et_models/onebody.jl
```diff
@@ -58,8 +58,13 @@ ___apply_onebody(selector, X::AbstractVector, E0s) =
          map(x -> E0s[selector(x)], X)
 
 
-site_grads(l::ETOneBody, X::ET.ETGraph, ps, st) = 
-         fill(VState(), (ET.maxneigs(X), ET.nnodes(X), ))
+# ETOneBody energy only depends on atom types (categorical), not positions.
+# Gradient w.r.t. positions is always zero.
+# Return NamedTuple matching Zygote gradient structure with empty edge_data.
+# The calling code checks isempty(∂G.edge_data) and returns zero forces/virial.
+function site_grads(l::ETOneBody, X::ET.ETGraph, ps, st)
```
cortner, line 64: I don't think this will lead to type stable code. If you look at the gradients that are produced by the other model components it will be of quite a different form: there will be additional fields in the returned NamedTuple and the edge_data field will contain VState or NamedTuple and not PState variables. I think this may need a bit of iteration.

That said, it might be a bit much hoping for perfect type stability here. I'll follow up on Slack on how to proceed.

src/et_models/stackedcalc.jl
```diff
@@ -0,0 +1,219 @@
...
+   end
+end
+
+# ============================================================================
+#  Efficient implementations using @generated for compile-time unrolling
+# ============================================================================
+
+@generated function _stacked_energy(sys::AbstractSystem, calc::StackedCalculator{N}) where {N}
+   quote
+      @nexprs $N i -> E_i = AtomsCalculators.potential_energy(sys, calc.calcs[i])
```
cortner, line 65: This refers to my other comment about the OneBody gradient. This is exactly the correct way to ensure type stability because you have N expressions for N calculators rather than a for loop. So I don't think there will be a type stability problem anywhere.

src/et_models/onebody.jl
cortner: I think the correct code might be 
```julia 
return (; edge_data = fill(VState(), length(X.edge_data)))
```
The reason this should work is that 
```julia
VState(r = SA[1.0, 2.0, 3.0]) + VState() == VState(r = SA[1.0, 2.0, 3.0])
```
i.e. an empty `VState()` just acts like zero when added on any other `VState`.

jameskermode:
Thanks for the feedback! Addressed both items:

1. **ET compat → 0.4.3** - Updated in both `Project.toml` and `test/Project.toml`

2. **`site_grads` type stability** - Changed to return `fill(VState(), length(X.edge_data))` instead of an empty array. As you noted, empty `VState()` acts as additive identity so this works correctly when summed with gradients from other model components.

All tests pass (946 passed, 1 broken - the known Julia 1.12 hash ordering issue).

src/et_models/et_calculators.jl
```diff
@@ -0,0 +1,784 @@
...
+- `st` - Model state
+- `rcut::Float64` - Cutoff radius for graph construction (Å)
+- `co_ps` - Optional committee parameters for uncertainty quantification
+"""
+mutable struct WrappedSiteCalculator{M, PS, ST}
+   model::M
+   ps::PS
+   st::ST
+   rcut::Float64
+   co_ps::Any
```
cortner, line 70: - `co_ps` currently not supported, but can leave as placeholder for the future.
- to discuss how to best implement this without breaking GPU memory barriers
- how important is type stability for exporting? The `Any` will break it. OTOH if it is unimportant, then `model, ps, st` could all be `Any` themselves, which will give more flexibility e.g. allow us updating `st` with preallocated memory etc. calculator then acts as a function barrier to keep type instability at the outermost layer.

None of these comments are blocking.

src/et_models/et_calculators.jl
```diff
@@ -0,0 +1,784 @@
...
+   # Handle empty edge case (e.g., ETOneBody with small cutoff)
+   if isempty(∂G.edge_data)
+      return zeros(SVector{3, Float64}, length(sys))
+   end
+   # forces_from_edge_grads returns +∇E, negate for forces
+   return -ET.Atoms.forces_from_edge_grads(sys, G, ∂G.edge_data)
+end
+
+# Compute virial tensor from edge gradients
+function _compute_virial(G::ET.ETGraph, ∂G)
```
cortner, line 98: ok here for now, but I like the idea of moving this into the ET atoms extension, because it is basically the pullback of the graph construction w.r.t. the cell vectors.

[non-blocking]

src/et_models/et_calculators.jl
```diff
@@ -0,0 +1,784 @@
...
+# Example
+```julia
+calc = ETACEPotential(et_model, ps, st, 5.5)
+E = potential_energy(sys, calc)
+```
+"""
+const ETACEPotential{MOD<:ETACE, PS, ST} = WrappedSiteCalculator{MOD, PS, ST}
+
+# Constructor: creates WrappedSiteCalculator with ETACE model directly
+function ETACEPotential(model::ETACE, ps, st, rcut::Real)
```
cortner, line 192: unclear to me why this is needed or useful at all?

cortner: same for any `ETACEPotential` specialized methods below.

cortner: Continuing reading, I see this is done for all three model components. I sort of see why it's needed, but I think there is a huge amount of code that can be unified here. 

Thinking about it, I think we should ignore my concerns about this for now, then think about how to move the core information - namely the mapping of center-species -> Dofs indices - to the inner model, and rewrite this basis interface accordingly. 

[non-blocking]

test/et_models/test_et_calculators.jl
```diff
@@ -0,0 +1,937 @@
...
+for s in model.pairbasis.splines
+   s.itp.itp.coefs[:] *= 0
+end
+
+# Convert to ETACE model
+et_model = ETM.convert2et(model)
+et_ps, et_st = LuxCore.setup(MersenneTwister(1234), et_model)
+
+# Match parameters
+et_ps.rembed.post.W[:, :, 1] = ps.rbasis.Wnlq[:, :, 1, 1]
```
cortner: I think you have some nicer implementation of the parameter conversion somewhere in `et_calculators.jl`?

test/et_models/test_et_calculators.jl
```diff
@@ -0,0 +1,937 @@
...
+# Test individual methods
+E = AtomsCalculators.potential_energy(sys, et_calc)
+F = AtomsCalculators.forces(sys, et_calc)
+V = AtomsCalculators.virial(sys, et_calc)
+
+@test E isa typeof(1.0u"eV")
+@test eltype(F) <: StaticArrays.SVector
+@test V isa StaticArrays.SMatrix
+
+println("AtomsCalculators interface: OK")
```
cortner: these println lines are useless. They will print regardless of whether or not the tests pass.

test/et_models/test_et_calculators.jl
```diff
@@ -0,0 +1,937 @@
...
+##
+
+@info("Testing cutoff_radius function")
+
+@test ETM.cutoff_radius(et_calc) == rcut * u"Å"
+println("Cutoff radius: OK")
+
+##
+
+@info("Performance comparison: ETACE vs original ACE model")
```
cortner: I generally prefer to not have performance benchmarks part of the CI runs. I do occasionally leave some simple benchmarks in the tests because it can be useful for interactive development, but I (try to) always comment them out before committing.

cortner (approved):
a few comments, none are crucial to be addressed. If we want to accelerate a bit we could just go ahead and merge but record those comments in an issue to be addressed over time.

Importantly, tests look ok to me as a first iteration and maybe the point when we really want to properly re-design tests and everything else is when we drop all the old codes entirely.

test/et_models/test_et_calculators.jl
jameskermode: done

test/et_models/test_et_calculators.jl
jameskermode: done

test/et_models/test_et_calculators.jl
jameskermode: yes, switched to that

jameskermode:
Thanks @cortner for the review. I've done the following:

- Removed println status messages from test file that provided misleading output
- Deleted ~100 lines of embedded CPU/GPU benchmarks from test file (benchmarks belong in benchmark/, not test/)
- Replaced manual parameter copying with utility functions copy_ace_params! and copy_pair_params!
- Made these parameter copying functions public (removed _ prefix) since they're used outside the module
- Added ET Calculators test to runtests.jl so it runs in CI
- Fixed `ETOneBody.site_grads` to return empty array `VState[]` instead of array filled with empty VStates - this was causing a FieldError when computing forces

I also opened issues #318 and #319 to keep track of the non-blocking comments. I think we can merge this PR and tag a new patch release now.

jameskermode:
(Just trying to fix the documentation CI failure, hold off merging for a few mins!)

cortner:
ping me when ready to merge

jameskermode:
Fixed now, happy for you to merge and then to merge the main co/etback PR too when you're ready. Export PR not too far away either.

Issue #314, opened 2025-12-23
Development Priorities


cortner:
### Linear models
   - [ ] get splines working with ET backend
   - [ ] introduce new pair potential model via V2rep + V2smooth (not compatible with old ACE1 style models)

### Non-linear models
   - [ ] Fully non-linear ACE fits - end-to-end Zygote-based AD through Lux-based ACE models
   - [ ] Pre-trained/exported embeddings (e.g. from MACE) then used for linear ACE models or TRACE models
  - [ ] TRACE

### General maintainance
 - [ ] Deprecate then drop support for pre-Lux/ET models
 - [ ] Bring back purification

(carried over from EquivariantTensors.jl)

PR #316, opened 2025-12-27
ET Pair Potential Model 

cortner:
Implements a pair potential model, compatible with ET and GPU evaluation, and the requirements for ACE models.


cortner:
This was a much bigger pain than expected. Dealing with two separate transformations x -> y -> polys and x -> r -> env was more complex than I thought and should maybe be avoided in the future. I added a note to #314

PR #323, opened 2026-04-11
Compatibility for PrettyTables v3

Alexsp32:
I think it’s pretty funny that a table printing package of all things is causing dependency issues for me when installing ACEpotentials alongside other packages. 

Here’s a fix which uses the newest v3 methods, but keep the existing code available for older versions.


cortner:
sorry I totally dropped the ball on this. For the future James is allowed to merge and tag on this repo. 

I think though that this is not really a breaking change so I'll regisgter it with a patch increment.

PR #326, opened 2026-06-29
Fix ~13x force-evaluation regression in classic ACEModel

jameskermode:
## Summary

The default `ace1_model` / classic `ACEModel` **force evaluation regressed ~13×** in the v0.10 (EquivariantTensors) series versus the previous release (0.9.1). This PR fixes it and adds regression tooling.

### Root cause
A **type instability** in `evaluate_ed` (`src/models/ace.jl`). `EquivariantTensors.pullback` is not type-stable (returns `Tuple{Any,Any}`), so `∂Rnl`/`∂Ylm` reached the inline gradient-assembly loops as `Any`, making every element-wise product dynamically dispatched. The numerical kernels were ~30 µs, but the call took ~400 µs and allocated ~764 KB/site — ~80% of the time was in the assembly loops.

The originally-suspected cause (ET **autograd** forces) was **refuted by benchmark**: the ET path is not the regression; in fact the broken classic path was slower than ET.

### Fix
Move the gradient assembly into a **function barrier** (`_assemble_grad_ed!`) so Julia re-specialises on the concrete runtime types.

| | before | after |
|---|---|---|
| `evaluate_ed` (40 neighbours) | ~386 µs | ~30 µs |
| classic forces, 512 atoms | 190.8 ms | 18.3 ms |
| cross-version vs 0.9.1 (512 atoms) | ~13× slower | ~1.15× |

Forces are **bit-identical** (max\|Δg\| ≈ 1e-15).

### Regression tooling (prevent recurrence)
- `benchmark/common.jl` — shared model/system setup
- `benchmark/bench_forces_regression.jl` — classic vs ET forces/energy (time + allocations)
- `benchmark/bench_crossversion.jl` (+ worker) — release-vs-release comparison in isolated envs
- `benchmark/profile_classic_ed.jl` — `evaluate_ed` component profile
- `benchmark/benchmarks.jl` — PkgBenchmark `SUITE`
- `.github/workflows/Benchmark.yml` — non-blocking BenchmarkCI judge (PR vs base on same runner, threads pinned)
- `benchmark/README.md`, `benchmark/FORCE_REGRESSION_FINDINGS.md`

### Verification
`test/models/test_ace.jl` and `test/models/test_calculator.jl` pass (finite-difference force checks, Zygote-through-loss).

### Follow-up (not in this PR)
Optimisations to the ET path (analytic `site_grads` to remove 12–154 MB/eval Zygote allocations; shared interaction graph across stacked calculators) — to be planned separately.


jameskermode:
@cortner OK to merge now or should we wait and fix in ET? My opinion is the function barrier does no harm and makes it work with current ET.

cortner:
this is definitely ok for me to merge, this would be ok like this even if we fix the type instability in ET

jameskermode:
I'll plan to merge this and #327 tag a patch release to get the regression fix to users

PR #327, opened 2026-06-29
ET path: share interaction graph + analytic pair site_grads

jameskermode:
Optimises the ET (EquivariantTensors) backend force path. Found by profiling (`benchmark/profile_et_forces.jl`).

> ~~**Stacked on #326** — base is `fix/classic-force-evaluation-regression` so the diff shows only the ET changes. Retarget to `main` once #326 merges.~~ #326 is now merged, retarged this PR on `main`

## Changes

**1. Shared interaction graph in `StackedCalculator`.** Each stacked component (onebody, pair, many-body) previously rebuilt its own interaction graph per call. A per-cutoff cache (`gcache`, keyed on each calculator's own `rcut`) is now threaded through the stacked energy/forces/virial/efv calls, so components that share a cutoff (pair + many-body) build the graph once. Keying on each calculator's **own** cutoff keeps per-component cutoffs exact (no single-graph approximation); non-`WrappedSiteCalculator` components fall back to the plain `AtomsCalculators` interface. `_wrapped_*` gain graph-accepting methods.

**2. Analytic `site_grads` for `ETPairModel`** (replaces `Zygote.gradient`). For the pair model `site_basis_jacobian` returns `∂𝔹 == ∂R` directly (the pair basis is a linear sum over neighbours), so contracting it with the readout weights is exactly the per-edge gradient — no jacobian blow-up, and low coupling (uses only `site_basis_jacobian`/`rev_reshape_embedding`).

**Not changed:** the many-body `ETACE.site_grads` stays on Zygote. An analytic VJP via `ET._ka_pullback` was prototyped but only ~10–15% faster (ET's many-body kernel intermediates, not Zygote overhead, dominate the cost) and coupled too tightly to EquivariantTensors internals — not worth it.

## Results (single-thread, Si/O)

| full ET stacked forces | before | after |
|---|---|---|
| 64 atoms | 6.95 ms | 5.06 ms (~27%) |
| 256 atoms | 24.9 ms | 20.0 ms (~19%) |
| 800 atoms | 71.4 ms | 61.5 ms (~14%) |

ET/classic force ratio 2.31 → 2.06. ET remains ~2× slower than the classic path on CPU — its dominant cost is ET's many-body kernels, so the larger remaining win is upstream (leaner kernels / GPU), not in ACEpotentials.

## Verification
- `test/et_models/test_et_calculators.jl` (ET↔classic forces/virial/energy `<1e-6`, incl. StackedCalculator), `test/etmodels/test_etace.jl`, `test_etpair.jl` — all pass.
- Forces bit-consistent with the classic calculator.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ scratch
 .vscode
 .ipynb_checkpoints
 .CondaPkg
+
+# PkgBenchmark tuning artifact
+benchmark/tune.json
--- a/benchmark/bench_forces_regression.jl
+++ b/benchmark/bench_forces_regression.jl
@@ -84,13 +84,13 @@ function run_table()
 end
 
 # ---------------------------------------------------------------------------
-# Isolation: time the gradient step itself (Zygote `site_grads`) vs the
-# analytic `site_basis_jacobian`, with the graph precomputed — isolates the AD
-# cost from graph construction and force assembly.
+# Isolation: time the ETACE gradient step `site_grads` vs the full basis jacobian
+# `site_basis_jacobian`, with the graph precomputed — isolates the gradient cost
+# from graph construction and force assembly.
 # ---------------------------------------------------------------------------
 function run_isolation()
    println("\nIsolation: gradient step only (graph precomputed)")
-   println("| Atoms | site_grads (Zygote, ms) | site_basis_jacobian (analytic, ms) | ratio |")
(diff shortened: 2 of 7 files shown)


jameskermode:
Currently only shares graphs where the cutoff is equal – which is the case for the pair and many-body graphs in the performance tests above - could also compute graph at max(rcut) and filter for the smaller cutoff ones. Open to suggestions on whether this is worth the extra book-keeping.

